PR #7, opened 2015-06-16
changed badge

alexvandesande:


frozeman:
Why you re-introduced this?
Without changing the styles, you will break the layout as it applies the same badge style for the submenu to the main menu button. (remember here the badge is the text below the button
screen shot 2015-06-17 at 09 48 49
)

frozeman:
fixed already in the other PR

PR #8, opened 2015-06-17
fixed shadows for web view, added padding for non full frame

alexvandesande:


frozeman:
Why giving the shadow only 1px, this looks on a non retina display not like a shadow, but like a 1px grey border.
screen shot 2015-06-17 at 09 44 29

I prefer 5 px, it looks more smooth, and i'm also not a fan of the `-1px` on the left :)

PR #14, opened 2015-06-18
Browser Bar

alexvandesande:


frozeman:
Very nice. I add the functionality to the back and reload button.

We have to think about a solution for when the url text is to long (or the window to small)

Im not perferctly convinced by the disco URL loading, maybe we can add a progress bar , which will always take 5sec (animation from width 0 -> 100%) and then when its ready loaded we increase the width very fats to 100%. Like current browsers do it, this gives the impression of a loading progress bar.

frozeman:
Btw thanks for the nice separation of the animations, is way better this way.

I found a bug, where the slide-out button, doesn't work anymore on non selected tabs. i fixed it.

additionally i removed the animation of the little triangle when it gets moved to the left side on full tabs, as the triangle was moving all the way from the right to the left, which looked kind of odd. Now it just appears there, whichs makes more sense.

Please take a look.

PR #25, opened 2015-06-19
Transitional animations

alexvandesande:


frozeman:
Could you next time provide a short description of what you PR does, so i can faster know for what to look ;)

frozeman:
I fixed a minor issue in the breadcrumbs, when the url is empty. And also the reloading on start (by fixing my template-var package).

Additionally we should no show the "normal" url on load again, as this creates a odd flicker when you switch e.g. chats in the whisper dapp.

frozeman:
Otherwise great work :+1:

PR #33, opened 2015-06-24
Updated README.md to reflect the addition of submodules support.

bobsummerwill:
Updated README.md to reflect the addition of submodules support.  Added sudo to the electron install instructions, because this is a global node installation.

### DIFF
--- a/README.md
+++ b/README.md
@@ -11,13 +11,11 @@ To run mist in development you need [Node.js NPM](https://nodejs.org) and [Meteo
 
 Then install electron:
 
-    $ npm install -g electron-prebuilt
+    $ sudo npm install -g electron-prebuilt
 
 Start the mist UI interface first using Meteor:
 
-    $ cd mist/interface/public
-    $ git clone https://github.com/ethereum/dapp-styles
-    $ cd ..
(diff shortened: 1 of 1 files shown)


frozeman:
Please remove the sudo, as it is no necessary when you setup you system correctly ;)

Additional you should add then `$ git submodule update --init` step.

Thanks a  lot for your addition

bobsummerwill:
Please could you clarify "correctly"?   On my Mac, without the sudo it is no dice for that kind of global npm install.

I will add the submodules command for sure.

bobsummerwill:
OK, @frozeman.   Good to go!   Please do let me know what "correct" means for OSX.   Thank you.

frozeman:
You need to set the permission of `/usr/local/lib/node_modules` to your user. Then you never need sudo again.

frozeman:
Thanks :+1:

bobsummerwill:
np.   What is the magic mantra?   I will get that working and then add that to the README too.

I am trying ... but not quite right:

sudo chown -R bob /usr/local/lib/node_modules

frozeman:
should work. I used path finder for that.

PR #34, opened 2015-06-25
Ipc integration

frozeman:
Creates the IPC bridge between the ui and guest content to `geth`.
It will filter incoming requests and disallow admin endpoints for guests content.


obscuren:
I suggest you start squashing some of those commits. You're polluting the git history

frozeman:
the smaller commits on the end, where additions i made and needed to leave the office. so i rather commit than risking loosing them, when i walk around with my laptop.
But most commits are actually larger changes. I always try to make one commit per task.

But thanks for you're feedback :baby_chick:

obscuren:
Know your git :-) Squash commits next time. There's no point in keeping that kind of history.

1. Hack
2. commit many times
3. Hack
4. commit
5. `git rebase -i HEAD~#`

frozeman:
Looks good, thanks!




Liebe GrüßeFabian

PR #73, opened 2015-12-03
Update README.md

genecyber:


README.md
```diff
@@ -20,8 +20,8 @@ Once a Mist version is released the Meteor frontend part is bundled using `meteo
 
 To run mist in development you need [Node.js NPM](https://nodejs.org) and [Meteor](https://www.meteor.com/install) installed. When this is done, install Electron:
 
-    $ npm install -g electron-prebuilt
     $ curl https://install.meteor.com/ | sh
+    $ sudo npm install -g electron-prebuilt
```
frozeman: If you use sudo than you configured your mac wrong ;)
http://stackoverflow.com/questions/16151018/npm-throws-error-without-sudo

README.md
```diff
@@ -45,7 +45,7 @@ Switch the `global.mode` to `"mist"` in the `main.js`.
 
 For development we start the interface with a Meteor server for autoreload etc. Start the interface in a separate terminal window:
 
-    $ cd interface && meteor
+    $ cd mist/interface && meteor
```
frozeman: This was assuming you're already in the mist folder, because the previous steps went in there. Not sure if we would need to show this in each step again. Otherwise i would need to do this in the others as well.. :\

frozeman:
Would close this for now, but i think the Readme over again. Thanks for the hints!

frozeman:
See here 64421b2

genecyber:
thanks

README.md
genecyber: good point, and I thought about that as I wrote that. It's common that -g flag requires sudo and the way to make -g install to local user space seemed... well not -g(lobal) 
That being said, I agree. It's too easy to set up users incorrectly and leave un necessary vulnerabilities that can be pushed to a prod env.

README.md
genecyber: ```Start the interface in a separate terminal window``` left me in ~/ But I was being nitpicky

Honestly I was expecting to to more edits but I ran out of day.

PR #101, opened 2016-01-06
page for git package installation

NikVolf:
Hello
merge If you are interested in this feature
idea is that user can just specify git repository on the special install page, it will be then downloaded locally and added to the left menu

can be tested on dummy repository
https://github.com/NikVolf/mist-hello-world

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,3 +3,5 @@ packages/
 interface_build/
 dist_wallet/
 dist_mist/
+.idea
+.npmrc
--- a/interface/client/appStart.js
+++ b/interface/client/appStart.js
@@ -23,7 +23,7 @@ mistInit = function(){
                 url: 'about:blank',
                 position: 0
             });
-            
+
             Tabs.insert({
                 url: 'http://ethereum-dapp-wallet.meteor.com',
                 position: 1,
@@ -31,6 +31,14 @@ mistInit = function(){
                     accounts: web3.eth.accounts
                 }
             });
+        }
+
+        if(!Tabs.findOne("install")) {
+            Tabs.insert({
+                _id: "install",
+                position: 9999,
+                name: "Install"
+            })
         }
     }, 1500);
 
--- a/interface/client/styles/elements.import.less
+++ b/interface/client/styles/elements.import.less
@@ -24,3 +24,11 @@ img {
     font-size: 0.9em;
     color: @colorTextSecondary;
 }
+
+input:invalid, input:focus:invalid {
+    background-color: rgb(255, 110, 110);
+}
+
+input[type='submit'] {
+    width: 100px;
+}
--- a/interface/client/styles/layout.import.less
+++ b/interface/client/styles/layout.import.less
@@ -154,3 +154,8 @@ webview, div.webview-content {
 div.webview-content {
     margin: 60px 30px;
 }
+
+input:required:invalid, input:focus:invalid {
+    /* insert your own styles for invalid form input */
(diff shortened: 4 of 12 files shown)


frozeman:
Interesting idea.

there are some minor details in your implementation but otherwise well done.
i have to discuss with the team about this idea, because the main idea is to install apps via swarm.
@alexvandesande what you think?

Additionally should the package be installed in the `~/Library/Ethereum/App..support/Mist` folder, rather than the application package. 

The question here would be, how to make sure you port these installed apps when you move to a new computer etc.

But definitely a good idea for the time beeing.

frozeman:
I just tried to run it, but because i need to install nodegit i had to update my node.js version.

now when starting i get the following error:
```
Uncaught Exception:
Error: Module version mismatch. Expected 47, got 46.
    at Error (native)
    at process.module.(anonymous function) [as dlopen] (ATOM_SHELL_ASAR.js:140:20)
    at Object.Module._extensions..node (module.js:450:18)
    at Object.module.(anonymous function) [as .node] (ATOM_SHELL_ASAR.js:140:20)
    at Module.load (module.js:356:32)
    at Function.Module._load (module.js:313:12)
    at Module.require (module.js:366:17)
    at require (module.js:385:17)
    at Object.<anonymous> (/Users/frozeman/Sites/_ethereum/mist/node_modules/nodegit/lib/nodegit.js:8:12)
    at Module._compile (module.js:425:26)
```

As of this line `(/Users/frozeman/Sites/_ethereum/mist/node_modules/nodegit/lib/nodegit.js:8:12)` i assume he can't load nodgit. Any idea?

NikVolf:
yeah I had troubles with electron-nodegit integration as well
I believe you need to install electron-rebuild

npm install electron-rebuild

then reinstall nodegit and recompile electron this way:

./node_modules/.bin/electron-rebuild

as written here:
https://github.com/atom/electron/blob/master/docs/tutorial/using-native-node-modules.md

frozeman:
Thats not good, we can't have such dependencies, because it would make it impossible for people to work with mist develop versions..

NikVolf:
Well as far as I read that's common scenario for native modules in electron/nwjs
Nodegit and others do not publish precompiled version for all engines
I will look more carefully to this issue, maybe there is more simple solution

frozeman:
Until now i and no need for a native module :)
What about https://github.com/jprichardson/node-github-download

NikVolf:
Yeah I think that will do
no need for some rocket-size git libraries anyway
Should have looked into it myself, thank you
will update soon

frozeman:
There is also a security concern. If the dapps are loaded using the file:// protocol they might be able to access the localstore/indexedDB of the mist interface, which keeps the permissions.

NikVolf:
Security concern is pretty valid, yes - if localstorage is used unencrypted.
You were going to address it, right? Or all apps are supposed to depend on some http server? Sorry if I miss some point somewhere.

I have a naive thought of simple solution with custom protocol scheme (a-la <b>"dapp://<APPID HERE>/index.html"</b> with hooks to forbid out-of-directory access since app id is given) - will it work? Or you have some more advanced solution?

frozeman:
I thought about the same, writing a custom protocol, but the teste i made didn't work right away for meteor applications, as they expected HTTP (and complex js apps are a good way to test this, rather than a simple html page).

Normally no app should be loaded on the file protocol, rather than mist. And it should be forbidden, based on the websecurity flag.

I made some comments on the last two commits, please check. Also merge develop again please.

PR #103, opened 2016-01-07
page for git package installation

NikVolf:
Hello
merge If you are interested in this feature
idea is that user can just specify git repository on the special install page, it will be then downloaded locally and added to the left menu

some discussion in the original commit to the wrong branch:
https://github.com/ethereum/mist/pull/101

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,3 +3,5 @@ packages/
 interface_build/
 dist_wallet/
 dist_mist/
+.idea
+.npmrc
--- a/README.md
+++ b/README.md
@@ -5,6 +5,8 @@ therefore it might not run out of the box as described the the README below.
 To save us time getting you all up and running, better wait for now. Thanks :)
 
 [![Join the chat at https://gitter.im/ethereum/mist](https://badges.gitter.im/Join%20Chat.svg)](https://gitter.im/ethereum/mist?utm_source=badge&utm_medium=badge&utm_campaign=pr-badge&utm_content=badge)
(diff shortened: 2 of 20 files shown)


NikVolf:
Fabian, about this note 
<i>"I thought about the same, writing a custom protocol, but the test i made didn't work right away for meteor applications, as they expected HTTP (and complex js apps are a good way to test this, rather than a simple html page)"</i>

How do you expect meteor applications to work at all? I believe they rely on the node process to serve requests from javascript application and allow node code to be ran by application? Isn't allowing applications to run code in node context even less secure?

What if we don't use custom protocol, but instead run special node process for apps? Generally it will be just a dedicated node process which will serve static files with root in the application folder.

Like 
<b>http://localhost:34599/index.html</b>

And it will just serve files under <b>~/Mist/Applications/==APPLICATION ID HERE==</b>

With this approach there is a way to make Meteor application work if we instead of regular node.js process run dedicated meteor process, but it should be somehow isolated to restrict code from application to access anything except the content of it's folder.

frozeman:
The meteor apps are static files, bundled with [meteor-build-client[(https://github.com/frozeman/meteor-build-client), no meteor server necessary...

I planned on creating a custom protocol like `dapp://mydapp/index.html`, which will serve the dapps from disk.

frozeman:
There is already a file called, customProtocol.js, in the mist folder, if you want to take a look.

frozeman:
Why closing it? I would have kept it open, until i would get to it.

NikVolf:
Ah no problem
I have read some about ipfs/swarm,
and started to consider this way to distribute apps to be a little naive

alexvandesande:
This functionality is being replaced by swarm in the future. Thanks @NikVolf

PR #342, opened 2016-03-13
Linux config location

Aranad:
I noticed the location for the Mist config for Linux in the ReadMe.txt has a question mark after it, so I set it to ~/.config/Mist but not sure if that applies to non-Debian-based systems too...


frozeman:
Thanks.

PR #361, opened 2016-03-15
Mist ux Improvements

alexvandesande:
* Moves the node info to the sidebar
* Adds a meta tag capture to detect the url bar style
* Adds css to support that new meta tag


interface/client/mistAPIBackend.js
```diff
@@ -56,6 +56,13 @@ mistAPIBackend = function(event) {
         }});
     }
 
+    // SET APPBAR
+    if(event.channel === 'appBar') {
+        Tabs.update(template.data._id, {$set:{
+            appBar: arg
```
frozeman: To what is it setting it? a boolean value?

interface/client/templates/elements/nodeInfo.html
```diff
@@ -27,7 +27,8 @@
         {{else}}
             <i class="icon-feed"></i> {{TemplateVar.get "peerCount"}} {{i18n 'mist.nodeInfo.peers'}}
             &nbsp;&nbsp;|&nbsp;&nbsp;
-            <i class="icon-layers"></i> {{formattedBlockNumber}} <i class="icon-clock" style="margin-left: 10px;"></i> {{{timeSinceBlock}}}
+            <i class="icon-layers"></i> {{formattedBlockNumber}} 
+            <br> <i class="icon-clock" style="margin-left: 10px;"></i> {{{timeSinceBlock}}}
```
frozeman: This marginl-left might be removed i guess.. It looked a bit weird in the screen sharing when i saw it.

interface/client/templates/layout/browserBar.html
```diff
@@ -1,5 +1,5 @@
 <template name="layout_browserBar">
-    <div class="browser-bar {{#if TemplateVar.getFrom currentWebView 'loading'}}loading{{/if}}">
+    <div class="browser-bar {{#if TemplateVar.getFrom currentWebView 'loading'}}loading{{/if}} {{#if appBar}}url-bar-{{appBar}}{{/if}}">
```
frozeman: If you move the {{with dapp}} above this bar, then the appbar is right away available and you don't need to create that extra helper..

interface/client/templates/layout/browserBar.js
```diff
@@ -75,6 +75,18 @@ Template['layout_browserBar'].helpers({
     */
     'currentWebView': function(){
         return '.tab-view webview[data-id="'+ LocalStore.get('selectedTab') +'"]';
+    },
+    /**
+    Return the app bar style
+
+    @method (appBarTransparent)
```
frozeman: See the previous comment.

interface/client/mistAPIBackend.js
alexvandesande: to whatever value the user adds on their `ethereum-dapp-url-bar-style` meta tag.

interface/client/templates/elements/nodeInfo.html
alexvandesande: True. This was just a 'get out of the way' change, I intend to do more work on that particular piece of css in the future, specially so it resizes more elegantly when there's more content

PR #392, opened 2016-03-19
gnat Wallet

valerii1969:


frozeman:
We only merge into master, once we release and a description with your intention would be helpful ;)

Issue #264, opened 2016-03-06
Look for a system wide installation of geth or eth


5chdn:
I'm maintaining the mist packages for ArchLinux and need some guidance.

Starting up mist always tries to start a geth node in `${pkgdir}/usr/share/${pkgname}/resources/node/geth/geth`, but usually people install and maintain their own version of geth or eth in their system, e.g., `/usr/bin/geth`...

Would it be possible that mist checks whether any suitable client is available in `$PATH` before starting it's own node?

Discussion from [AUR](https://aur.archlinux.org/packages/mist/):

> For some odd reason /usr/share/mist/resources/node/geth/geth has no execution privileges after installation with pacman. This is why I'm getting the same error as @notungx when I run mist. Adding the privilege solves the problem.
> 
> The file is not marked as executable probably because it's unpacked from a .zip achive, which has no concept of access privileges. I see that PKGBUILD explicitly sets privileges for libnode.so and Ethereum-Wallet. It should be doing that to geth as well. Please add this line at the end of package():
> 
> ```
> chmod 755 "${pkgdir}/usr/share/${pkgname}/resources/node/geth/geth"
> ```
> 
> For a moment I thought that maybe lack of exacutable flag is deliberate - after all geth is an optional dependency here. But no - mist tries to use its own copy of geth even if you install the geth package from AUR. It does not seem to detect /usr/bin/geth. The dependency is misleading and should be removed.
> 
> The error and the dependency make it look as if geth wasn't bundled with mist until very recently. Is this what happened here? 

And:

> I get the following error after executing mist:
> 
> ```
> CONNECT to IPC PATH: /home/santi/.ethereum/geth.ipc
> CONNECT to IPC PATH: /home/santi/.ethereum/geth.ipc
> NODECONNECTOR ERROR { [Error: connect ECONNREFUSED /home/santi/.ethereum/geth.ipc]
> code: 'ECONNREFUSED',
> errno: 'ECONNREFUSED',
> syscall: 'connect',
> address: '/home/santi/.ethereum/geth.ipc' }
> Node type: geth
> Network: main
> Start node from /usr/share/mist/resources/node/geth/geth
> Stopping nodes...
> Starting geth node...
> UNCAUGHT EXCEPTION { [Error: spawn EACCES] code: 'EACCES', errno: 'EACCES', syscall: 'spawn' }
> Stopping nodes...
> ```
> 
> In my laptop works just executing mist without the need of executing geth first... any ideas?

[...]

PR #393, opened 2016-03-19
Added --help, --version and --mode cli parameters

hiddentao:
```
Usage:
mist [options]

Options:
  --version   Display app version
```

Note that on my system I can get this work even in dev mode, i.e. I get `path/to/Electron . [options]`.


hiddentao:
Need to do a bit more work on this for OS X. In progress.

hiddentao:
Working on Mac now:

```js
/Applications$ Mist.app/Contents/MacOS/Electron --version
0.5.2
```

hiddentao:
Added `--mode` parameter that allows you to set app model to either `mist` or `wallet` without having to edit main.js

5chdn:
Cool, I want to have something like `--client /path/to/geth`, is that possible? ref #264 :-)

hiddentao:
@5chdn Yep that would be easy to put in. Perhaps we can call it `--geth` instead.

I'm guessing the logic could be:  if `--geth` provided then always use that version of geth. If not found then error (i.e. don't use the bundled fallback).

frozeman:
@hiddentao did you test this with the bundled version on mac OS X as well?
Concerning the geth flag, i would rename this to `gethpath` and `ethpath`.

And i think it should not store this as the default path, as mist ships his own version on purpose, as its the version its optimised for. If people choose to set their own version, than they do this at a risk of incompatibility and can create a symlink, or script etc..

hiddentao:
@frozeman Yep, you can use the command-line options with the bundled version as long as you run it from the terminal, as illustrated in my previous comment above.

As for my comment that Mist should _always use that version of geth_ supplied on the command-line I should have made that clearer. I meant that for that running instance of Mist it should only use that supplied version of geth. I agree with you that Mist should by default always use it's bundled version unless `gethpath` is supplied as an option.

5chdn:
Isn't mist client-agnostic? So `--ethpath` or `--gethpath` would be limiting to two clients. `--client` could point to any executable which starts an ethereum node with ipc socket.

frozeman:
@5chdn thats already the case, if mist detects a running node on the default pic path, it uses this one... So the path flag wouldn't change a thing here. It would only be useful, if you want mist to start that node for you.

Eth and geth have a slightly different way to start, therefore we would need this separation for now. Otherwise i would agree that `--nodepath` is better...

hiddentao:
So just to be clear, we want to add `--ethpath` and `--gethpath`. i can go ahead and work on that. Having said that, If we could merge this in as it is as a first step that would be great.

hiddentao:
Added `--gethpath` and `--ethpath` parameters

hiddentao:
Replaced by #443

PR #437, opened 2016-03-30
Fix #416 - crash after adding new account

### LINKED ISSUE #416: Crash when creating new account
Node type: geth/eth
OS: linux mint
Network type: test
Mist version: 0.5.2

Ethereum Wallet crashes, when I go to menu -> accounts -> new account; after i repeated the new password and hit OK. 

attached you can find my log-files
[node.log.zip](https://github.com/ethereum/mist/files/187604/node.log.zip)

PS: it's my first use of this software, and the main window remains blank white after startup. only the menu bar is visible. is this normal?

hiddentao:
Fixes #416. This replaces underscore with lodash (as this provides for improved functionality) and fixes the crash which occurs after adding a new account.


hiddentao:
Ah, sorry I'd accidentally merged #393 into my `master` If we're not going to merge that one in I can redo this PR against official `master`.

frozeman:
I will take a look, but please always make PRs against develop only

README.md
```diff
@@ -37,7 +37,7 @@ Requires electron version 0.37.2
 To run mist in development you need [Node.js NPM](https://nodejs.org) and [Meteor](https://www.meteor.com/install) and electron installed:
 
     $ curl https://install.meteor.com/ | sh
-    $ npm install -g electron-prebuilt
+    $ npm install -g electron-prebuilt@0.36.9
```
frozeman: we are already at 0.37.2

modules/ipcCommunicator.js
```diff
@@ -4,6 +4,7 @@ Window communication
 @module ipcCommunicator
 */
 
+const _ = require('lodash');
```
frozeman: There are differences in the API, are you sure it doesn't break anything?
I mostly use `_.each` and `_.isEmpty`, which i saw exist as well.

My personal opinion is, yes it does have performance improvements, but for the use cases were we use it, it makes 0 difference..

modules/ipcCommunicator.js
```diff
@@ -34,7 +35,7 @@ ipc.on('backendAction_closeApp', function() {
 ipc.on('backendAction_closePopupWindow', function(e) {
     var windowId = e.sender.getId();
 
-    if(global.windows[windowId]) {
+    if(_.get(global.windows, windowId)) {
```
frozeman: I didn't know about `_.get()` seems like a neat function. Though for this use case its probably a bit to much ;)

hiddentao:
@frozeman All good points. Initially the code rewrite looked different, which is why I had put lodash in. Then I refactored it and though it doesn't really require lodash like you said I decided to leave it in since it's a better lib. I tried to ensure I replaced all usage instances of underscore with lodash, but without tests it's hard to say.

All in all, I think I'll redo this PR from scratch against `develop`. I should also then redo the other PR I have pending against `develop`.

hiddentao:
Replaced by #442

PR #442, opened 2016-03-31
Fix #416 - wallet crashes after adding account.

### LINKED ISSUE #416: Crash when creating new account

hiddentao:
This replaces #437

### DIFF
--- a/modules/ipcCommunicator.js
+++ b/modules/ipcCommunicator.js
@@ -9,6 +9,8 @@ const appMenu = require('./menuItems');
 const popupWindow = require('./popupWindow.js');
 const ipc = require('electron').ipcMain;
 
+const _ = global._;
+
 /*
 
 // windows including webviews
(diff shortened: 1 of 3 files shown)


modules/ipcCommunicator.js
```diff
@@ -32,32 +34,38 @@ ipc.on('backendAction_closeApp', function() {
...
 ipc.on('backendAction_sendToOwner', function(e, error, value) {
-    var windowId = e.sender.getId();
+    var windowId = e.sender.getId(),
+        senderWindow = global.windows[windowId];
 
-    if(global.windows[windowId]) {
-        global.windows[windowId].owner.send('windowMessage', global.windows[windowId].type, error, value);
+    var mainWindow = global.mainWindow;
 
+    if (_.get(senderWindow, 'owner')) {
```
frozeman: Are you're `_.get()` exists in underscore?
Im fine using lodash if you think its a better fit, as long as it also supports the api of underscore

utils/underscore.js
```diff
@@ -0,0 +1,33 @@
...
+  /**
+   * Get a deeply nested object property.
+   *
+   * @param {Object} obj The object.
+   * @param {String} path The path within the object to fetch.
+   * @param {*} fallbackValue The value to return if given path not found.
+   *
+   * @return {*} Returns value if found; otherwise the fallbackVAlue.
+   */
+  get: function(obj, path, fallbackValue) {
```
frozeman: Ah i see you implemented it.... hm maybe its just better to use loads then...

hiddentao: Like you said, it may break some other stuff (e.g. I had to alias a couple of lodash methods to ensure minimongo kept working). I think at some point we should definitely switch to lodash though I'm not yet familiar with enough of the codebase to say with certainty that it won't break something. Perhaps you have a better idea of this.

frozeman: Yeah. the minimongo could break, which i rather don't want to try out... haha

frozeman:
Thanks for the fix!

PR #443, opened 2016-03-31
Command-line arguments

hiddentao:
Replaces #393 

```bash
Usage:
mist [options]

Options:
  --version   Display app version
  --gethpath  Path to geth executable to use instead of default
  --ethpath   Path to eth executable to use instead of default
  -m, --mode  App mode: wallet, mist (default)
```


hiddentao:
Updated help usage with `getpath` and `ethpath` options.

frozeman:
People used the `mist --ignoreblacklistgpu` flag or so on ubuntu to fix their white screen issue, are you sure this would still work with this command line additions? A lot of people depend on this.

hiddentao:
Technically it should be fine since I don't modify the incoming argument list. I simply parse it. On the other hand what we could do is add this flag to the "usage" documentation so that people know they can use it. Also, did some searching around and it seems that if we want to be certain then we could turn the flag on manually within the code itself:

```js
app.commandLine.appendSwitch('ignore-gpu-blacklist', 'true');
```

frozeman:
Yes, but the problem is that enabling that by default can lead to unknown crashes, due to the blacklisting (which happened for a reason i guess)

hiddentao:
Fair enough. In that case I recommend we at least document it in the auto-generated help section (I can do that commit).

frozeman:
Yes please and ideally we test before if it for sure doesn't break the `ignore-cpu-blacklist` flag, or we need to pass it through ourself.

hiddentao:
Ok, I've added the `--ignore-gpu-blacklist` flag to the command-line - and just in case, I call `app.commandLine.appendSwitch('ignore-gpu-blacklist', 'true');` to ensure it really is set.

frozeman:
How do i use this in dev mode, when i didn't bundle the app yet?

5chdn:
Thanks!

Issue #472, opened 2016-04-06
_.mixin is not a function


lazukars:
Node type: geth
OS: Ubuntu 14.04
Network type: main
Mist version: 0.0.0 ( On master branch. Pulled 04-05-16 )

I've followed the install instructions here, https://github.com/ethereum/mist, to a tee. When I run `electron .`, the following error occurs. Please see the screenshot.

![error__mist](https://cloud.githubusercontent.com/assets/250052/14302720/fed390d4-fb71-11e5-9128-2978f06488c8.png)

hiddentao:
I just tried this on the latest `master` and didn't see this error. I get the GUI showing up and sync begins.

lazukars:
Hmm. Just tried pulling master again. I got the same issue. What OS did you try this on?

hiddentao:
OS X El Capitan 10.11.2

frozeman:
Can you try the `wallet` branch? that should have fixed it

lazukars:
So it looks like this was an issue with electron. I was running the wrong version. 

However, when running `npm install -g electron-prebuilt@0.37.2`, as instructions state, the error in the attached  screenshot occurs. 
![electron-bug](https://cloud.githubusercontent.com/assets/250052/14624399/5ccaa7a0-05a8-11e6-8173-6805e7f23228.png)

Note: About a month ago, I had none of these issues, and Mist worked fine.

hiddentao:
Could you try running `npm cache clean` and then again doing it? It looks like a possibly corrupted `package.json` in one of Electron's depenencies.

hiddentao:
Also, `npm uninstall -g electron-prebuilt` followed the installation call again.

lazukars:
I've done the following:


`npm uninstall -g electron-prebuilt`
`npm cache clean` which fixed that last screenshot error. 
`npm install -g electron-prebuilt@0.37.2`
`electron -v` : output: v0.37.2

`cd mist`
`git pull && git submodule update` ( on master branch )
`npm install`
`cd interface && meteor`

Now, when running `electron .` inside the mist dir, I'm getting the initial issue error again, _.mixin is not a function. See initial screenshot. I've tried `electron . --mode wallet` as well. The same _.mixin is not a function occurs.

hiddentao:
Could you try it on the `develop` branch please - that's always the actively developed branch. Just want to rule out branch differences.

lazukars:
[...]

PR #503, opened 2016-04-19
Client-side logging capture

hiddentao:
This captures `console.log` and related calls in the Mist UI layer and sends them via IPC to the backend log4js logging system. 

This PR also fixes a minor bug related to loading and playing UI sounds client-side.

**Note:** The Mist UI currently crashes me after loading due to some other unknown bug (other users are seeing it too - #472). Perhaps I can fix it as part of this PR.


hiddentao:
Output looks something like:

```
[2016-04-19 17:11:47.617] [INFO] checkNodeSync - No sync necessary, starting app!
[2016-04-19 17:11:47.618] [INFO] main - Loading Interface at http://localhost:3000
(electron) getUrl is deprecated. Use getURL instead.
[2016-04-19 17:11:48.679] [DEBUG] (ui: mist) - (ipc callback from window with url: http://localhost:3000/)
[2016-04-19 17:11:48.679] [INFO] (ui: mist) - Meteor starting up...
[2016-04-19 17:11:48.987] [INFO] (ui: mist) - http://about:blank
```

interface/client/appStart.js
```diff
@@ -41,6 +39,27 @@ mistInit = function(){
 
 
 Meteor.startup(function(){
+    // send console logging to IPC backend
```
frozeman: Thats exactly what should be as a module in a preloaded file ;)so that not every dapp needs to have that and we can watch all dapps logs, if we wanted to

interface/client/mistAPIBackend.js
```diff
@@ -17,15 +19,7 @@ var filterId = function(str) {
 };
 
 
-var sound = {
-    bip: document.createElement('audio'),
-    bloop: document.createElement('audio'),
-    invite: document.createElement('audio'),
-};
-sound.bip.src = 'file://'+ dirname + '/../../sounds/bip.mp3';
```
frozeman: I did it like this, so that the sounds are preloaded. With your approach it would load the file in the moment it uses them, which could lead to delays, i guess.

modules/ipc/ipcProviderBackend.js
```diff
@@ -593,4 +597,24 @@ module.exports = function(){
     ipc.on('ipcProvider-writeSync', function(event, payload){
         sendRequest(event, payload, true);
     });
-};
\ No newline at end of file
+
+    const uiLoggers = {};
+
+    ipc.on('console_log', function(event, id, logLevel, logItemsStr) {
```
frozeman: This whole stuff belongs not here, but in: https://github.com/ethereum/mist/blob/master/modules/ipcCommunicator.js

interface/client/mistAPIBackend.js
hiddentao: This was throwing errors in the JS console on Mist startup, because `dirname` is not available in the Meteor client-side. Neither is `__dirname` or any of the standard Node.js stuff, though Meteor 1.3 probably has support for it. In any case, since the file is being loaded from the local FS delays should be minimal.

interface/client/appStart.js
hiddentao: Yeah, good idea.

modules/ipc/ipcProviderBackend.js
hiddentao: Will move.

interface/client/mistAPIBackend.js
frozeman: dirname is added in the preloaded file for the mist interface. Though it might be smart to move change that do using `_dirname` as its anyway a module loaded in the preloaded file
Its a custom global variable i added. See: https://github.com/ethereum/mist/blob/develop/modules/preloader/mistUI.js#L33

hiddentao: That's interesting. I definitely have some problem with the preloader then - this may also be what caused my previous crash.

hiddentao:
Shifted the code into a preloader now and changed where IPC listens to it. Much nicer now.

hiddentao:
@bas-vk @alexvandesande Would be nice to get another pair of eyes on this.

interface/i18n/app.de.i18n.json
```diff
@@ -1,34 +0,0 @@
...
-        "you": "Du",
-        "send": "Senden",
-        "or": "oder",
-        "of": "von",
-        "with": "mit",
-        "and": "und",
-        "on": "auf",
-        "per": "pro"
-    }
-}
```
alexvandesande: Why did you delete these german translation files?

alexvandesande:
I tried running your branch but got this error from electron:

```
App threw an error when running [SyntaxError: Block-scoped declarations (let, const, function, class) not yet supported outside strict mode]
```
I ran `npm install` to check if it was a new dependency but the error persisted.

hiddentao:
What version of node are you using?

interface/i18n/app.de.i18n.json
hiddentao: Sorry about that, fixed now.

alexvandesande:
@hiddentao v0.12.3

hiddentao:
Ah ok, so Node 4 is required for this. That's the current LTS release version of node and allows us to use the latest ES6 features. Will update package.json and the README.

Issue #79, opened 2015-12-08
Small typo: "Sharing a Identity"


tcoulter:
From here: https://github.com/ethereum/mist/blob/fc9d8d9f756f902a07b98f3e84e8f72e4c7b3b98/interface/client/templates/layout/browserBar_accounts.html#L4

Should be "sharing **an** identity".

alexvandesande:
Thanks for all the issues Tim!

alexvandesande:
Thanks, merged

PR #513, opened 2016-04-24
fixes typo

luclu:
see #79


luclu:
will redo into `develop`

PR #548, opened 2016-05-03
Fix Mist startup crash due to about:blank page

hiddentao:
This fixes the crashes experienced by some OS X users (see #472). 

1. Change `about:blank` to `ethereum.org` for default browser tab in Mist. 
2. Add command-line option to reset Mist interface tabs - a command-line option is required since an in-app GUI option won't be usable if the app keeps crashing.

### DIFF
--- a/interface/client/appStart.js
+++ b/interface/client/appStart.js
@@ -15,19 +15,25 @@ The init function of Mist
 @method mistInit
 */
 mistInit = function(){
-
+    console.log('Initialise Mist');
 
     Meteor.setTimeout(function() {
+        if (0 <= location.search.indexOf('reset-tabs')) {
+            console.log('Resetting UI tabs');
+            
+            Tabs.remove({});
+        }
+
         if(!Tabs.findOne('browser')) {
             Tabs.insert({
                 _id: 'browser',
-                url: 'about:blank',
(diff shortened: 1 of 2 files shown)


alexvandesande:
Same error I had with your other PR unfortunately. Probably my fault thou..

interface/client/appStart.js
```diff
@@ -15,19 +15,25 @@ The init function of Mist
...
+            console.log('Resetting UI tabs');
+            
+            Tabs.remove({});
+        }
+
         if(!Tabs.findOne('browser')) {
             Tabs.insert({
                 _id: 'browser',
-                url: 'about:blank',
+                url: 'https://ethereum.org',
```
alexvandesande: We will change this to the app catalog when we have one

hiddentao:
Error is due to node version. My changes require that you are running Node 4 or above. This is the currently LTS release of Node anyway so we should be using it.

luclu:
@hiddentao should we update the readme including the node >= 4 req?

hiddentao:
@luclu Yep that's a good idea. I can do this in #559

hiddentao:
Actually looks like I'd already put it in, see https://github.com/ethereum/mist/tree/develop README

PR #559, opened 2016-05-04
Node startup and synchronization improvements

hiddentao:
I'm heavily refactoring the code which starts up geth and syncs to it. This is intended to both make the code easier to understand and to reduce the frequency of sync errors we seem to be getting with it.

* Decoupled the UI from the connection and sync logic - using event-driven architecture instead. This will make stuff more easy to test now and in future.
* Create a dedicated `Socket` class for managing socket connections more intelligently
* Refactor `ethereumNodes` to more intelligently start, shutdown and manage the main Geth node IPC connection
* Rewrote synchronisation logic (see nodeSync.js) to be less error-prone, more intelligent.
* Refactor how we create and manage windows in the Electron layer so that we avoid errors occurring due to a window already having been destroyed
* Splash screen logic simplified. It's now much more in tune with what's going on under the hood. We also now show which network you're on in the top-right to reduce confusion. "MAIN-NET", "TEST-NET" and "DEV-NET".
* We have a "Goto app" button that's available as soon as IPC is connected (this replaces the previous "Skip peer search" button).
* Switching between test and live network whilst running the app should be less painful now. Note that there is still some IPC connection logic which needs to be refactored for a truly painless experience. 
* Much more logging throughout the code to help diagnose issues in future
* Updated connection logic code to handle Geth 1.4 
* Improved the fix for #548 
* Upgraded to work with Electron 1.0.1


hiddentao:
I think this is ready for review now. There's still a bit more work I can do from the UI side of things to improve sync. But I thought it might be a good idea to review what's been done so far and make sure it works. Also, once Geth 1.4 I want to make sure these changes work with it.

hiddentao:
Now that Geth 1.4 is out I think I should test it with these changes to make sure everything works. It also seems that users see sync stopping once the main UI screen is loaded. I haven't yet gotten to that part of the "sync" code, perhaps I can take a look at that too.

hiddentao:
@bas-vk @alexvandesande @frozeman @luclu This PR is ready for review.

interface/client/templates/elements/networkIndicator.html
```diff
@@ -0,0 +1,19 @@
+<template name="elements_networkIndicator">
+    <div class="network-indicator">
+        {{#if TemplateVar.get "unknown"}}
+            <div class="unknown indicator" type="Network not yet known">CHECKING NETWORK...</div>
```
frozeman: Should be {{i18n...

interface/client/templates/elements/networkIndicator.js
```diff
@@ -0,0 +1,82 @@
...
+
+/**
+Check network type.
+
+@method checkNetworkType
+*/
+var checkNetworkType = function(template) {
+    console.trace('Check network type...');
+
+    try {
```
frozeman: Why the try block? It should return an error in the callback if something is wrong

interface/client/templates/elements/networkIndicator.js
```diff
@@ -0,0 +1,82 @@
...
+
+
+Template['elements_networkIndicator'].onRendered(function(){
+    var template = this;
+
+    TemplateVar.set(template, 'unknown', true);
+
+    checkNetworkType(template);
+
+    ipc.on('nodeStatus', function(e, status) {
```
frozeman: I would prefix the ipc messages like I did in other files. To prevent overlaps in the future. 
Eg use `uiAction_...` or so

hiddentao: Done.

interface/client/templates/elements/networkIndicator.js
hiddentao: There is no callback. What happens is that the UI says _"checking network..."_. It will keep saying this until this call is successful.

interface/client/templates/elements/networkIndicator.html
hiddentao: Done.

hiddentao:
Have noticed that if I do `brew install ethereum` the auto-started geth instance has its datadir set to `/usr/local/Cellar/ethereum/1.4.4/.ethereum` rather than in `~/Library/Ethereum`. This is why Mist can’t connect via IPC to the default Geth - which means it tries to start its own, which will obviously fail.

zackkoGH:
Wallet needs to be synched before ANYTHING can be done with the wallet itself.

How difficult would it be to have the wallet loading and synching IN BACKGROUND like Bitcoin Core works?

I'm using Ethereum as an hobbyist and synching the wallet once a week: it takes more than 90 minutes to sync every time on a Fibre connection of more than 30Mb/s speed. 
This is very much not user-friendly.

Thank You.

hiddentao:
@zackkoGH This PR will allow you to go straight to the app as soon as the sync process begin - you won't have to wait anymore :)

alexvandesande:
I'm doing a PR on this PR to fix some issues on the template mechanism

zackkoGH:
@hiddentao Thank You for the update.
Will look forward for when this feature will be release.

alexvandesande:
@hiddentao please see https://github.com/hiddentao/mist/pull/1/

hiddentao:
Thanks.

-- 
Ramesh Nair

Director
HiddenTao Ltd (UK company no. 6807289)

http://hiddentao.com
https://github.com/hiddentao
https://www.linkedin.com/in/hiddentao


On 20 May 2016 at 07:30:31, Alex Van de Sande (notifications@github.com) wrote:

Merged #559.

PR #587, opened 2016-05-09
Update french translation of tap-i18n.json

olivierstern:


alexvandesande:
Thanks for the corrections Olivier, I'm closing this PR without merging because you're editing the correct files in other threads already

PR #612, opened 2016-05-12
Translation in italian

h2002al:
Proof of translation in Italian.


alexvandesande:
Thanks. Will merge when you have also added these files:
https://github.com/ethereum/mist/blob/develop/interface/i18n/mist.en.i18n.json

PR #613, opened 2016-05-12
Translation in Albanian

h2002al:
Proof of translation in Albanian,


alexvandesande:
Thanks. Please also translate this file:

PR #618, opened 2016-05-12
Added Dutch translation files for the interface

actuallymentor:
Translated based on my knowledge as a Dutch native.

Some notes:
- I used the formal 'You' (u instead of jij)
- Some words can be translated to Dutch, but natives tend to use the English, so I left some in English
- Added two commas to the English language file as well
- Exact phrasings from Dutch windows settings may be slightly different, I have no Dutch windows machine to check


alexvandesande:
Thanks! If you want the actual wallet files to be translated as well, please do so here:

actuallymentor:
I'll try and spend some time on those as well, though it might take me some days with my current schedule.

alexvandesande:
Thanks for finishing it

PR #626, opened 2016-05-13
Luclu fix 372

alexvandesande:


luclu:
thanks! will add the ones for the new languages, too.

PR #665, opened 2016-05-18
fixed linux32 node path and binary

luclu:
updated path according to gulp-file to include geth binary on linux32. Fixed linux-binary from karabale as current seems to be linux-arm.


hiddentao:
Is this ready for review? please mark it as such if so.

PR #673, opened 2016-05-20
Check for node startup error

hiddentao:
This will check to see if Geth fails to startup due to a fatal error, specifically if it fails to bind to the UDP port it needs. Checking for this explicitly allows us to handle this failure case properly + plus it fixes the network switching functionality.


alexvandesande:
Couldn't reproduce the error this is meant to fix, but I've tested it and runs with me, so the next 👍 can merge.

hiddentao:
Added a more helpful error message if Geth unable to bind to the required UDP port.  This happens often for users if Geth is already running (but on a different IPC path) and/or if it failed to shut down. This error msg will make things clearer.

To test this scenario, start Geth with IPC path set to something other than normal (this way Mist will try and start its own node). Then start Mist - it should show an error message saying that another instance node may be running.

luclu:
It looks good to me - tried several cases in OSX and linux.

PR #674, opened 2016-05-20
Remove old, redundant code

hiddentao:
Remove old, redundant code.


alexvandesande:
Aren't we using popupWindow.js for launching the new accounts page?

hiddentao:
I refactored that all into `modules/windows.js`.

alexvandesande:
Yes, tested and it works. Will wait for a bit for another 👍 and eventually merge it

luclu:
Sorry guys this PR/branch disabled the import-dialog for me. Also I don't get any onboarding screen on a vanilla osx (I moved `~/Library/Application Support/Mist` and `~/Library/Ethereum`).

I don't know that much about branching strategies, but the current process either is a mess or it is just confusing me (which is about as bad 😞).

hiddentao:
@luclu I wonder if some cacheing is involved in the issues you're having, thought I can't think where the cache might be.

alexvandesande:
I didn't have any of these problems Luca

PR #679, opened 2016-05-22
small typo: some_where

luclu:
fixes #678 

The automatically `pulblic/i18n` folder is - I think unecessarely tracked in git. I will create a follow-up PR on that.

PR #680, opened 2016-05-22
Untrack public/i18n/

luclu:
follow-up on #679

The automatically pulblic/i18n folder is - I think - unecessarely tracked in git.


hiddentao:
Agreed!

luclu:
I will look into meteor/gulp-automatisation to populate `public/i18n` a prior the packaging: https://github.com/ethereum/mist/issues/691

alexvandesande:
When you finish this, please apply the same solution to the wallet-dapp

luclu:
This PR without additional i18n processing by meteor will only bundle english strings.
Looking into this in: https://github.com/ethereum/mist/issues/691

frozeman:
Thats why it needs to run meteor first to produce the other i18n files. 
Bundling always worked before :)

we would need to add the step to the gulp file to run meteor before and quit it after a while. 

As this is quite annoying, it might just be better to keep the files in the repo....

luclu:
You are right. The best way I figured out - apart from that - is to replicate tapi18n-bundler's process in gulp...
`meteor build` has different usecases and can't be leveraged for this.

PR #677, opened 2016-05-20
Integrate #601 along with some improvements to IPC Path

alexvandesande:
Reopened https://github.com/ethereum/mist/pull/652 against develop

Issue #156, opened 2016-02-10
Add command-line parameters like --help, --version


gcarq:
Would be nice if we can add common CLI parameters like --help and --version.

5chdn:
i second that, would be a good idea to be able to pass command line arguments to the underlying ethereum node, too.

but that's a design decision that should be carefully discussed. i can not assess the impact on security for instance.

frozeman:
How you start mist using the command line?

gcarq:
`$ /usr/bin/mist`

which is symlinked against:
`-rwxr-xr-x 1 root root 61887528 Feb  9 16:42 /usr/share/mist/Ethereum-Wallet`

frozeman:
This package is not maintained by me. As this has to be cross OS compatible i have to look if its possible. 
Ill keep this as a reminder.

frozeman:
BTW the node is in `/usr/share/mist/Ethereum-Wallet/resoucres/node/geth/`, where you can simply start it before with your own options and then start the wallet, it will then use the running node.

hiddentao:
PR in #393

5chdn:
I like the existing flags as merged in https://github.com/ethereum/mist/pull/443 already. But I have some more on my wishlist.

    --ipcpath     # which ipc path to connect to (e.g., if you connect to
                  # a client with non standard parameters, ...)
    --datadir     # allow storing the blockchain in different locations
                  # (pass this to the client)
    --testnet     # run mist on testnet

ref. http://ethereum.stackexchange.com/q/3518/87

5chdn:
Or to add another thought, (and to simplify the approach), why not simply passing any flag from mist to geth/eth. For instance, if you run `mist --dev` it will run a client in dev mode. Just an idea.

gcarq:
@5chdn I like that idea. `mist --dev` could also be `mist --` to signify the end of *mist* options like bash does, and all following arguments are passed to the underlying node.

hiddentao:
PR in #689

luclu:
This will be available in version 0.7.5.

PR #689, opened 2016-05-23
Pass custom CLI options to the Ethereum node

hiddentao:
This PR fixes the remaining work in #156. You can now pass startup options to Geth/Eth as follows:

```bash
$ electron . --loglevel debug -- --rpcport 18343 --networkid2
```

Updated the usage instructions:

```bash
Usage:
electron . [Mist options] -- [Node options]

Mist options:
  --mode, -m              App mode: wallet, mist.     [string] [default: "mist"]
  --ipcpath               Path to node IPC socket file (this will automatically
                          get passed as an option to Geth).             [string]
  --gethpath              Path to Geth executable to use instead of default.
                                                                        [string]
  --ethpath               Path to Eth executable to use instead of default.
                                                                        [string]
  --ignore-gpu-blacklist  Ignores GPU blacklist (needed for some Linux
                          installations).                              [boolean]
  --reset-tabs            Reset Mist tabs to their default settings.   [boolean]
  --logfile               Logs will be written to this file in addition to the
                          console.                                      [string]
  --loglevel              Minimum logging threshold: trace (all logs), debug,
                          info, warn, error.          [string] [default: "info"]
  --version, -v           Display Mist version.                        [boolean]

Node options:
  -  All options will be passed onto the node (e.g. Geth).

Options:
```

This PR also merges #677 as the work overlapped.


alexvandesande:
Had the following error when tried to run this:
`App threw an error when running [SyntaxError: Identifier 'Settings' has already been declared]`

hiddentao:
Ah, minor bug - now fixed.

alexvandesande:
Now works for me

PR #696, opened 2016-05-23
Nav bar glitch and new art for Mist

alexvandesande:


hiddentao:
What's the nav bar glitch I need to test?

alexvandesande:
Sometimes you would put an URL and it would load that page but not update the url element and therefore you couldn't pin it to the sidebar. Basically I had to go through multiple URLs old and new until the bug reproduced but then it would happen all the time

luclu:
fyi your last commit added some log to the tree.

alexvandesande:
Thanks Luca, I removed it now

luclu:
Fixed it for me.

PR #705, opened 2016-05-25
Norwegian Bokmål translation

sandakersmann:
Norwegian Bokmål translation


alexvandesande:
Thanks. If you'd also like to see the wallet translated, you'd need to change these files:

Also add your language to these lists:

https://github.com/ethereum/meteor-dapp-wallet/blob/develop/app/project-tap.i18n

sandakersmann:
I will translate those as well when I find some time :)

PR #706, opened 2016-05-25
Remove dependency on node log output for detecting startup success

hiddentao:
As the title says. We now wait 4 seconds after starting the node - if no error is returned within that time then we assume the node has started up successfully and continue onto attempting the IPC connection.


modules/ethereumNode.js
```diff
@@ -526,19 +544,12 @@ class EthereumNode extends EventEmitter {
...
 
-                        if ('eth' === nodeType) {
-                            // (eth) prevent started until str appears
-                            if (-1 === dataStr.indexOf('jsonrpc')) {
-                                log.trace('Running eth so wait until we see JSONRPC message');
-
-                                return;
-                            }
-                        } else if ('geth' === nodeType) {
+                        if ('geth' === nodeType) {
```
frozeman: I find this way around confusing :)
better is: `nodeType === 'geth'`

or do you have a reason for that?

modules/ethereumNode.js
```diff
@@ -488,6 +488,22 @@ class EthereumNode extends EventEmitter {
 
                 const proc = spawn(binPath, args);
 
+                /*
+                    Assume startup succeeded after 5 seconds. At this point 
+                    IPC connections are usually possible.
+                */
+                setTimeout(() => {
```
frozeman: This should only be started, in `.stderr.on('data', ...)`, at least then we know for sure that it at least started

modules/ethereumNode.js
hiddentao: Defensive programming. If by mistake `=` is used it won't compile. Whereas if you did `nodeType = "geth"` that still compiles and will produce an undesired outcome.

modules/ethereumNode.js
hiddentao: Agreed. I've changed it so that as soon as any data is received (stdout or stderr) it starts the timer.

alexvandesande:
Worked for me without a glitch

PR #707, opened 2016-05-25
Check for update at startup and inform the user if a new version is available

hiddentao:
When the app runs this checks the ethereum/mist Github releases list for newer release that's not in draft mode. If one is available (newer release will have a version tag > version in app's package.json) then a popup will be shown to the user informing them of this. The internal logic checks to see what you're running (Mist or Wallet) and adjusts the look and feel and text accordingly.

The popup contains a link to the specific new release's page on Github releases. 

A "Check for updates..." item in the main menu (under "About") also allows the user to trigger this check at will.


luclu:
works for me!
But the icon looks a bit off to the side.
Could you also add a link to the releases page?

bildschirmfoto 2016-05-25 um 10 36 50

hiddentao:
Woah, this is still a work in progress ;) Will let you all know when it's ready.

hiddentao:
Ok, this is now ready for review! Here is how to test this:

**Test 1: When no new version is available**

1. Start Mist from command-line as normal
2. You should see normal behaviour
3. Once Mist has started up goto Menu -> Check for Updates...
4. This should show a popup window informing that you that it is checking for updates. 
5. After a second or so it should say that no updates were found.

Repeat the above procedure for the Wallet.


**Test 2: When a new version is available**

1. Edit `package.json` and change the version to 1.00
1. Start Mist from command-line as normal
2. You should see the update checker popup automatically show up informing you that an update is available. 
3. If you click the link in the popup it should take you to the page for the latest release.
4. If you close it you should be able to continue using the app as normal
5. You can still goto Menu -> Check for Updates... at any time to check for updates.

Repeat the above procedure for the Wallet.

luclu:
haha, guess I was to curious ;)
it works and looks good to me!
Only nothing happens when I click on the download-link.

bildschirmfoto 2016-05-25 um 11 10 25

hiddentao:
Sorry about that, now fixed.


On 25 May 2016 at 17:12:53, Luca Zeug (notifications@github.com) wrote:

haha, guess I was to curious ;)
it works and looks good to me!
Only nothing happens when I click on the download-link.

modules/updateChecker.js
```diff
@@ -0,0 +1,114 @@
...
+        case 'wallet':
+            str = 'wallet';
+            break;
+    }
+
+    return got('https://api.github.com/repos/ethereum/mist/releases', {
+        timeout: 3000,
+        json: true,
+    })
+    .then((res) => {
```
luclu: Would it be possible to add a differentiation between **Mist** and **Ethereum-Wallet** as otherwise users will get a prompt if we release another **Mist** version. And vise-versa for another **Etherum-Wallet**.

hiddentao: That was the intention (for the future) but I saw we haven't done a Mist version so merged them for now - maybe I'm wrong about that, is there a Mist version?

hiddentao: Updated now anyway.

modules/updateChecker.js
```diff
@@ -20,6 +20,8 @@ const check = exports.check = function() {
     switch (global.mode) {
         // right now we are just releasing the wallet!
         case 'mist':
+            str = 'mist':
```
luclu: typo :P

luclu: oh sorry, I meant the missing semicolon `str = 'mist':`

frozeman:
There is one released mist version.

interface/client/lib/helpers/templateHelpers.js
```diff
@@ -30,12 +30,49 @@ Template.registerHelper('CurrentBlock', function(){
...
+    return window.mist.mode === 'mist' ? 'Mist' : 'Ethereum Wallet';
+});
+
+/**
+Return the app icon path.
+
+@method (iconPath)
+**/
+Template.registerHelper('appIconPath', function(){
+    return 'file://'+ window.mist.dirname +'/icons/'+ window.mist.mode +'/icon2x.png';
```
frozeman: Why this PR is containing stuff alex added a while back, or did you change thet stuff @hiddentao?

interface/client/styles/popupWindows.import.less
```diff
@@ -9,6 +9,28 @@
...
+        h1 {
+            font-size: 200%;
+            margin-bottom: 20px;
+
+            &.no-upate {
+                font-size: 180%;
+            }
+        }
+
+        #image {
```
frozeman: id's are a no go :) especially if you nest it them in a class. i would simply target the image, or use a class

main.js
```diff
@@ -191,6 +196,9 @@ app.on('ready', function() {
     // Initialise window mgr
     Windows.init();
 
+    // check for update
+    require('./modules/updateChecker').run();
```
frozeman: Why are you checking for the update in the backend and not in the update check window?

I would argue that this just adds unnecessary complexity to the backend and extra pic connections, which might lead to memory leaks (i try to keep these kind of connections as less as possible.)

modules/updateChecker.js
```diff
@@ -0,0 +1,115 @@
...
+    switch (global.mode) {
+        case 'mist':
+            str = 'mist';
+            break;
+        case 'wallet':
+            str = 'wallet';
+            break;
+    }
+
+    return got('https://api.github.com/repos/ethereum/mist/releases', {
```
frozeman: Where is the `got` function from?

frozeman:
I didn't test, but looked over the code and it seems fine.

The only note i have is my first comment:
"id's are a no go :) especially if you nest it them in a class. i would simply target the image, or use a class"
If this was already there, i'm unclear why it was added.

interface/client/lib/helpers/templateHelpers.js
hiddentao: I changed things. I added in more global template helpers that made sense - and then removed the local equivalents from the "about" template. Because I needed the same ones in the "updateAvailable" template.

interface/client/styles/popupWindows.import.less
hiddentao: I left the ids as they were because I figured there was probably a good reason for them being there. Have now removed them.

modules/updateChecker.js
hiddentao: `fetch` isn't yet available in Node, it seems. So I opted for a lightweight wrapper around Node's `request`.

hiddentao:
Pending issues fixed.

alexvandesande:
I like how it detects Mist upgrades as well.

captura de tela 2016-05-26 as 12 18 07 pm

PR #708, opened 2016-05-25
Update geth nodes to 1.4.5

alexvandesande:
This among others, fixes the estimate gas function


luclu:
linux32 is missing

frozeman:
Linux32 we need to get from peter directly

PR #721, opened 2016-05-27
Fix Mist/Wallet app bundling

hiddentao:
This fixes an issue introduced by #688 which stopped Mist/Wallet from being bundled properly. 

This merges `cliArgs` into `Settings` and makes that the one-stop shop for all app settings (though we still write some to the `global` object - I will refactor this later).

The build break occurred because the gulp script previously had a dependency on specific lines of code within `main.js`. This has been refactored such that we now create a `config.json` file in the root folder during bundling which will determine the internal app settings. A much more robust solution.


alexvandesande:
While packaging I noticed that the icons were not being packaged anymore and instead it was using the default electron icon for mac. Diving into the issue it seems that the issue if with the electron version. When I change ```gulpfile.js``` line 39 from ```var electronVersion = '1.0.0';``` to ```var electronVersion = '0.37.2';``` it will deploy the icon correctly. 

Of course we'd like to keep the updated electron, but I haven't found yet the issue.

hiddentao:
@alexvandesande Just a note that Electron **1.0.1** is our current version, not 1.0.0.

hiddentao:
@alexvandesande I did the build and saw the Electron icon, same as you. I then copied the .app to my Applications folder at which point it was showing the correct Wallet icon. Bizarre! But perhaps this means it is actually working?

hiddentao:
Merged #727

alexvandesande:
@hiddentao that would be a weird behavior indeed. Are you sure you weren't looking at an older version of the app? I just did this and it remains with the default electron icon even on the application folder..

luclu:
It's working for me.

frozeman:
You can refresh the app icon using `$ touch path/to/electron.app`. The icon is cached to some degree.
I will look over your PR if i find time.

alexvandesande:
Since it's not caused by this PR I separated the icon issue in a different one: https://github.com/ethereum/mist/issues/762

Will merge this now

PR #724, opened 2016-05-27
Refactoring Dapp IPC connection

hiddentao:
Refactored the Dapp IPC connection logic to reduce the no. of errors and make it more robust.


hiddentao:
This is now ready.

hiddentao:
Finally it's ready!

alexvandesande:
I'm able to sync without any apparent issues. Will wait a bit before merge to see who else has these problems

modules/ipc/ipcProviderBackend.js
```diff
@@ -1,670 +1,400 @@
...
-        
-        } else if(event) {
-            event.returnValue = true;
+     * Handler for when Ethereum node state changes.
+     *
+     * Auto-reconnect sockets when ethereum node state changes
+     *
+     * @param {String} state The new state.
+     */
+    _onNodeStateChanged (state) {
```
frozeman: Thats problematic, as pub/sub connections won't be informed  and those are socket bound notifications.

What i do in the frontend is, reloading the dapps, which anyway creates a new socket

modules/ipc/ipcProviderBackend.js
```diff
@@ -1,670 +1,400 @@
...
-                } else {
-                    if(!event.sender.isDestroyed())
-                        event.sender.send('ipcProvider-data', JSON.stringify(result));
-                    delete _this.asyncEvents[event.eventId];
-                }
-            });
-        });
+        log.trace('sendRequest', isSync ? 'sync' : 'async', ownerId, payload);
 
+        let jsonPayload = null;
```
frozeman: are you sure, this is not creating memory leaks?
i always try to avoid using variables outside of functions, like you use this one here in the promise functions..

modules/ipc/ipcProviderBackend.js
```diff
@@ -1,670 +1,400 @@
...
-                if(event.sync)
-                    event.returnValue = JSON.stringify(returnValue(jsonPayload, returnVal));
-                else
-                    event.sender.send('ipcProvider-data', JSON.stringify(returnValue(jsonPayload, returnVal)));
+        let allResults = ([].concat(payload)).map((item, idx) => {
+            let ret = {
+                jsonrpc: "2.0"
+            };
 
+            if (value) {
```
frozeman: I assume this is what introduces the bug that sometimes results are returned without `.result` property!

Did you take batch RPC requests into account? which come in form of an array.
Or results, which are objects, or arrays, or simple strings?

modules/ipc/ipcProviderBackend.js
hiddentao: I don't follow. How can the DApp know that the node has disconnected?

modules/ipc/ipcProviderBackend.js
hiddentao: No, definitely not. I did quick test to confirm that GC works ok. I put the variable outside to make it shareable across the promise `then` handlers.

modules/ipc/ipcProviderBackend.js
hiddentao: I left it like that based on what I saw in the previous code, see https://github.com/ethereum/mist/blob/4908a106b1808bb148b1e6cb2eefbf64d7bc0a4e/modules/ipc/ipcProviderBackend.js#L62

PR #741, opened 2016-05-29
Upgrade to web3 version in order to fix broken send tx functionality

hiddentao:
Rather than wait for a new web3 version and keep code broken I've fixed the web3 dependency to the specific Git commit - https://github.com/ethereum/web3.js/commit/0f1ab3227b8985fa5b9cd1fdc75315383f4d7249


hiddentao:
Updated to `0.17.0-alpha` published by @frozeman

PR #768, opened 2016-06-02
Full translation in italian

AlessandroCrepaldi:
File has been translated in Italian.


PabloEscobarZ:
@AlessandroCrepaldi, ciao scusami, come faccio selezionare la lingua italiana?...

PR #772, opened 2016-06-03
Refactor splash screen

alexvandesande:
* Block numbers change every millisecond instead of every second

* Displays state downloading status as a ghost progress bar

* Centers the “checking network” message


hiddentao:
Looks good but how can I see state download bar?

PR #789, opened 2016-06-08
Fix problem with geth node being detached

hiddentao:
Fixes:

* If you exit Mist geth will also exit properly
* Socket IPC connector (for dapps) is now more intelligent and stable
* "Launch Application" link in splash screen shows sooner


luclu:
Works fine for me. I tested quiting Mist at different stages of it's lifecycle and network-switching.

Although, perhaps this should be rebased on https://github.com/ethereum/mist/pull/786 to avoid those "launch app link" things. Also getting that "Testnet"-string in the top-middle of the splash-screen, when using the test-net.
bildschirmfoto 2016-06-08 um 12 28 29

alexvandesande:
Works great for me, thanks Ram!

PR #792, opened 2016-06-08
Fix the testnet label alignment

alexvandesande:
* Fixes the alignment of the testnet label
* Adds more i18n strings
* Changes .floor to .round for the number animation


hiddentao:
Please label this PR.

hiddentao:
Looks good.

PR #796, opened 2016-06-09
Fix handling of array results in IPC calls to backend

hiddentao:
This fixes a number of bugs introduced by the DAPP IPC refactor (#724).

### DIFF
--- a/modules/ipc/ipcProviderBackend.js
+++ b/modules/ipc/ipcProviderBackend.js
@@ -378,16 +378,14 @@ class IpcProviderBackend {
     @method makeReturnValue
     */
     _makeReturnValue (originalPayload, value) {
-        value = [].concat(value);
+        let finalValue = _.isArray(originalPayload) ? value : [value];
 
         let allResults = ([].concat(originalPayload)).map((item, idx) => {
             let ret = {
                 jsonrpc: "2.0"
             };
 
-            if (value) {
-                ret.result = value[idx];
-            }
+            ret.result = finalValue[idx];
 
             ret.id = item.id;
             
--- a/modules/ipc/ipcProviderBackend.js
+++ b/modules/ipc/ipcProviderBackend.js
@@ -343,6 +343,10 @@ class IpcProviderBackend {
                 delete e.message;
             }
 
+            // delete stuff leftover from request
+            delete e.params;
+            delete e.method;
+
             e.id = item.id;
 
             return e;
--- a/modules/ipc/ipcProviderBackend.js
+++ b/modules/ipc/ipcProviderBackend.js
@@ -112,9 +112,9 @@ class IpcProviderBackend {
                     log.trace('Notification received', ownerId, data);
 
                     if (data.error) {
-                        data = this._makeError({}, data);
+                        data = this._makeErrorResponsePayload({}, data);
                     } else {
-                        data = this._makeReturnValue({}, data);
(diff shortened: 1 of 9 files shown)


hiddentao:
Todo:

* Share more validation between request and response phases
  * If admin mode then eth_accounts should not be filtered
  * "allowed methods" check should take place on responses too
* If a single request within a batch returns an error still return the other results

frozeman:
Conversation notes:

- move validators into processors
     - so that admins and other permission will be taken into account
- make sure that requests and responses, for batch and normal are filtered/validated
- make better error filter for batch requests: [req, notallowreg, req] -> sends actually -> [req, req] -> returns [result, error, result]
     - check batch requests
     - if throw doesn't interfere with them and invalidate all returns/requests

frozeman:
Ok, awesome, i edited your todos slightly

hiddentao:
Ok, just submitted a fix which hopefully covers the remaining issues. Did some brief testing on a private net.

modules/ipc/methods/base.js
```diff
@@ -14,38 +16,101 @@ module.exports = class BaseProcessor {
...
+            throw this.ERRORS.INVALID_PAYLOAD;
+        }
+
+        if (this._isAdminConnection(conn)) {
+            return;
+        }
+
+        // prevent dapps from acccesing admin endpoints
+        if(!/^eth_|^shh_|^net_|^web3_|^db_/.test(payload.method)){
+            delete payload.result;
```
frozeman: If the son RPC looks like:
```
--> {"jsonrpc": "2.0", "method": "subtract", "params": [42, 23], "id": 1}
<-- {"jsonrpc": "2.0", "result": 19, "id": 1}
```
If it has a `.method` it has no result object, that would only be the case for a request. So in this case you would add an error to the method?
If youre building an error object, you would also need to delete the `.method` i guess.

hiddentao: I add an `error` in the next line. Is deleting the `method` in case of error strictly necessary? I've assumed not.

modules/ipc/ipcProviderBackend.js
```diff
@@ -288,66 +294,43 @@ class IpcProviderBackend {
...
-        }
-
-
+    _sanitizeRequestPayload (conn, payload) {
         if (_.isArray(payload)) {
-            for (let p of payload) {
+            _.each(payload, (p) => {
                 if ('eth_sendTransaction' === p.method) {
-                    throw ERRORS.BATCH_TX_DENIED;                    
+                    p.error = ERRORS.BATCH_TX_DENIED;
```
frozeman: Adding the error object alone, is not enough, you would need to remove the `method` and `params` property too.

Maybe adding a clean result functions somewhere later?

frozeman:
Apart form these few questions, it looks good to me! Thanks ram!

hiddentao:
Fixed. Error responses no longer include method and params.

frozeman:
If meteor reloads the wallet interface, `web3.isConnected()` returns `false`, even though its connected.

This changed with your PR, can you check whats wrong here?

Its implemented here: https://github.com/ethereum/web3.js/blob/master/lib/web3/ipcprovider.js#L162-L170

Its definitely your PR, which changed it

alexvandesande:
Works for me. Will wait for @frozeman to merge.

PR #811, opened 2016-06-14
Upgrade to Electron 1.2.2

hiddentao:
This also fixes a minor Web3 connection issue (reported by @frozeman in #796)


frozeman:
How does the splash window perform on the new electron

hiddentao:
I don't notice any difference to be honest. I'm doing more for the automated testing ticket.

frozeman:
Ok i checked and it looks fine

PR #822, opened 2016-06-14
fixes issue where onboarding screen would not appear

alexvandesande:


main.js
```diff
@@ -309,8 +309,8 @@ app.on('ready', function() {
             .then(function getAccounts() {
                 return ethereumNode.send('eth_accounts', []);
             })
-            .then(function onboarding(result) {
-                if (ethereumNode.isGeth && result && result.length === 0) {
+            .then(function onboarding(resultData) {
```
frozeman: @hiddentao why you named that function?

frozeman:
Looks good! Thanks alex

main.js
hiddentao: If it throws an error then in the stack trace it will show up, rather than "anonymous function", etc. Plus it makes the code easer to read.

hiddentao:
Thnx Alex.

PR #825, opened 2016-06-15
Update eth to 1.2.9

alexvandesande:
Just uploading the new eth to the github repo. Will work in the future to add it to the automatic download, like geth.


luclu:
Mmh, there seems to be some ipc issue. Can't connect to node on OS X and windows 7.

PR #827, opened 2016-06-15
gulp: shasum not consistent with output string 'shasum -a 256'

luclu:


luclu:
@frozeman now it should be correct :)

PR #839, opened 2016-06-16
Fixed wallet tab insertion, added admin permission

frozeman:
This adds a permission property `admin: true`. When a tab has this it can see all accounts and do all admin like actions.

This also fixes the tab insertion. I keep this open to add more fixes as i go


interface/client/appStart.js
```diff
@@ -19,30 +19,26 @@ mistInit = function(){
 
     if (0 <= location.search.indexOf('reset-tabs')) {
         console.info('Resetting UI tabs');
-        
         Tabs.remove({});
     }
 
     if(!Tabs.findOne('browser')) {
-        console.debug('Insert tabs');
```
hiddentao: Why has this been removed?

modules/ethereumNode.js
```diff
@@ -218,12 +218,12 @@ class EthereumNode extends EventEmitter {
...
-                .then(() => {
-                    this.state = STATES.STOPPED;
-                    this._stopPromise = null;
-                });
+            .then(() => {
+                this.state = STATES.STOPPED;
+                this._stopPromise = null;
+            });
         } else {
-            log.debug('Disconnection already in progress, returning Promise.');
```
hiddentao: Why has this been renamed? It can be re-connected after disconnection, hence why it is called that.

frozeman:
Closing in favour of #911 911

PR #841, opened 2016-06-16
Fixes the node start IPC connection

frozeman:
This PR is adding a loop to test an IPC connection. When a connection timeout is set, it tries to connect every 200ms to the connection, if not it fails.

This speeds up the connection process, as we don't have to wait arbitrary 4 seconds anymore (which weren't enough for `eth`)

@hiddentao This PR works, but if you set a wrong IPC path it doesn't quit gracefully, instead i delivers an uncaught exception. Could you please look into this.
Simply let the `eth` node start (Change the `node` files content in `~/Library/App..Support/Mist/node` to `eth`) 
Before change the pic path in the `getIPCPath.js` to some other path.

You will see it fails, but not gracefully, as the error seems to travers through and doesn't get catched.

This error is created in the socket.js line 84

```js
[2016-06-16 16:44:28.827] [INFO] Sockets/node-ipc - Connect to {"path":"/Users/frozeman/Library/Ethereum/gethdddd.ipc"}
[2016-06-16 16:44:38.965] [ERROR] Sockets/node-ipc - { Error: connect ENOENT /Users/frozeman/Library/Ethereum/gethdddd.ipc
    at Object.exports._errnoException (util.js:949:11)
    at exports._exceptionWithHostPort (util.js:972:20)
    at PipeConnectWrap.afterConnect [as oncomplete] (net.js:1080:14)
  code: 'ENOENT',
  errno: 'ENOENT',
  syscall: 'connect',
  address: '/Users/frozeman/Library/Ethereum/gethdddd.ipc' }
Error: connect ENOENT /Users/frozeman/Library/Ethereum/gethdddd.ipc
    at Object.exports._errnoException (util.js:949:11)
    at exports._exceptionWithHostPort (util.js:972:20)
    at PipeConnectWrap.afterConnect [as oncomplete] (net.js:1080:14)
[2016-06-16 16:44:38.967] [ERROR] main - UNCAUGHT EXCEPTION { Error: connect ENOENT /Users/frozeman/Library/Ethereum/gethdddd.ipc
    at Object.exports._errnoException (util.js:949:11)
    at exports._exceptionWithHostPort (util.js:972:20)
    at PipeConnectWrap.afterConnect [as oncomplete] (net.js:1080:14)
  code: 'ENOENT',
  errno: 'ENOENT',
  syscall: 'connect',
  address: '/Users/frozeman/Library/Ethereum/gethdddd.ipc' }
[...]

PR #420, opened 2016-03-26
Possibility to specify IPC path & RPC over HTTP support

tomusdrw:
Based on #393 

1. With this PR it would be possible to specify path to ipc socket (`--rpc` cli option)
2. There is also mock version of `net.Socket` which uses HTTP as a transport.

I would really love to get rid of that: https://github.com/ethereum/mist/compare/develop...tomusdrw:rpc?expand=1#diff-a31b48952900046cc9945315da84b939R22
It seems that the responses are stuck somewhere if the requests are sent synchronously (`setImmediate` or lower timeouts are also not sufficient). But the response is genererated by server and also quiting the application causes the response to be received and processed (sic!). Any hints?

PR #871, opened 2016-06-22
Enable HTTP RPC sockets

hiddentao:
This incorporates and replaces ethereum/mist#420. 

* It replaces the `--ipcpath` CLI parameter with `--rpc`, allowing for an HTTP RPC hostname (e.g. `http://localhost:8545`) to be provided. If omitted then IPC is used as the default.
* README updated with instructions
* Introduce usage of `--node-` prefix for CLI parameters that are to be passed onto geth/eth.
* Warns user that the HTTP option is less secure than IPC
* Fix a small bug in solidity compilation error reporting
* Incorporates ethereum/mist#841 as I needed that to test this properly.


hiddentao:
Ready for review.

hiddentao:
Todo:

* Add warning pop-up

hiddentao:
Dialog msg added.

screen shot 2016-06-28 at 12 25 58

frozeman:
haha :)
Some text improvements:

"....This is less secure than using a local IPC socket - your account passwords will be send in plain text!

Only do this when you secured your HTTP connection, or you know what you are doing."

luclu:
@hiddentao could you also correct [this](https://github.com/ethereum/mist/blob/4922876722fcae21732ed634acb87beccbbaf232/modules/settings.js#L140) typo?

hiddentao:
@luclu Fixed directly in `develop` branch.

modules/settings.js
```diff
@@ -136,8 +137,13 @@ for (let optIdx in argv) {
 }
 
 // some options are shared
+<<<<<<< HEAD
```
frozeman: This should be removed

frozeman:
There seems to be an issue, if i put `web3.eth.accounts` into the interface it doesn't return, and mist wont quit.
Then sometimes when quitting, it will show all the RPC messages, because it can't parse it properly.

I will look more deeply what could be the problem, but if you have any idea @hiddentao please let me know.

modules/settings.js
hiddentao: Done.

hiddentao:
@frozeman Yep, haven't been able to reproduce it yet. If I see it happen I'll investigate further.

hiddentao:
This is ready to be merged for 0.8.2.

Issue #833, opened 2016-06-16
rename "Ethereum-Wallet" to "Ethereum Wallet"


luclu:
PR https://github.com/ethereum/mist/pull/883

alexvandesande:
Have you tested this PR on multiple systems?

luclu:
Done.

Issue #881, opened 2016-06-24
OSX app consistent naming


sammy007:
Could you please finally decide how to name OSX app?

![screen shot 2016-06-24 at 19 41 10](https://cloud.githubusercontent.com/assets/7374093/16340874/c4082a76-3a43-11e6-966f-305276795975.png)

luclu:
Ah, sorry for this. The name hasn't been changed in the codebase - due to some windows specific problem and didn't make it into this release due to different releasers not being aware. 
We will push this into `0.7.7` and also switch our packaging process to leverage travisCI soon.

I will close this in favour of https://github.com/ethereum/mist/issues/833.

sammy007:
Ah thanks, didn't notice it.

luclu:
PR in https://github.com/ethereum/mist/pull/883

PR #883, opened 2016-06-24
change binary name "Ethereum-Wallet" to "Ethereum Wallet"

luclu:
`0.7.5` used **Ethereum Wallet** while
`0.7.6` used **Ethereum-Wallet** again
as reported in https://github.com/ethereum/mist/issues/881 (OSX app consistent naming)

This PR closes https://github.com/ethereum/mist/issues/833.

Tested on OSX 10.11.5, Ubuntu 14.04 and Windows 7 x64.

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -113,7 +113,7 @@ gulp.task('set-variables-mist', function () {
 gulp.task('set-variables-wallet', function () {
     type = 'wallet';
     filenameLowercase = 'ethereum-wallet';
-    filenameUppercase = 'Ethereum-Wallet';
+    filenameUppercase = 'Ethereum Wallet';
     applicationName = 'Ethereum Wallet';
 });
 


gulpfile.js
```diff
@@ -113,7 +113,7 @@ gulp.task('set-variables-mist', function () {
 gulp.task('set-variables-wallet', function () {
     type = 'wallet';
     filenameLowercase = 'ethereum-wallet';
```
luclu: What is this?

evertonfraga: idk... it isn't used across the entire project.

alexvandesande:
Previously I was changing the filenames manually. This change you made also would impact on the folder structure and windows platform. Not that simple

gulpfile.js
hiddentao: That code is used as part of the build process.

hiddentao:
Seems to me that it's just the Mac app folder name which needs to be set. Am I right?

gulpfile.js
frozeman: This was used in an earlier stage of the build script. we can probably get rid of it now

frozeman:
Chasing the file name will most likely have effects on other places. @luclu did you build the files and see if it all works?

luclu:
I build and diffed all the files and folders.
The only changes happen to the binary name **and** the platform's folder name.
Both get set during electron-packager's `name` variable [gulpfile.js#L301](https://github.com/ethereum/mist/blob/develop/gulpfile.js#L301).

frozeman:
can you remove the `filenameLowercase` then

luclu:
I will try to find another variable of electron-packager to differentiate the folder from the binary name. Otherwise we might have to leverage gulp to rename it afterwards.

frozeman:
Yeah the folder name and possible problems with some OS, or servers is why i kept the dash.
I'm not sure if introducing a lot of complicity makes sense here.

If we go that route than rather renaming the file names afterwards. The problem is renaming in gulp is not straight forward, as you need to copy, rename and delete the old one. I do this quite a lot in the current gulp script already.

So i vote for less complexity.

luclu:
I will try to leverage https://www.npmjs.com/package/gulp-rename - haha funny nodejs world 60.000 download in the last day..

frozeman:
we already use `gulp-rename`, but like i said it doesn't remove the old version, so you need to have a second step removing it yourself. It just adds complexity imo

luclu:
On hold until https://github.com/ethereum/mist/pull/972 is merged.

luclu:
Done in PR https://github.com/ethereum/mist/pull/1159 (former #972).

Issue #888, opened 2016-06-25
cleanup i18n language codes


luclu:
**Rationale:**
Stick to language code standards to:
 - prevent broken subsystems/incompatibility
 - prevent consistency throughout different components (Mist, Dapps, etc.)
 - provide clear guidelines for translators

---

**Proposal:**
Current most accepted code standard is RFC 4646: language in lowercase (ISO 639-1),
then a hyphen, then the region in uppercase (ISO 3166-1 alpha-2). ([ref: *Camilo Sperberg*, 2013, i18n*:Things to take into account](http://unreal4u.com/talks/?get=EN-internationalization.pdf))

[List of (**RFC 4646**) language codes](https://github.com/TAPevents/tap-i18n/blob/master/lib/plugin/etc/language_names.js), excerpt:
```
af	    Afrikaans
af-ZA	Afrikaans (South Africa)
ar	    Arabic
ar-AE	Arabic (U.A.E.)
ar-BH	Arabic (Bahrain)
ar-DZ	Arabic (Algeria)
ar-EG	Arabic (Egypt)
[...]
```

luclu:
Currently supported languages **|** current code **|** RFC4646 code

luclu:
done in https://github.com/ethereum/mist/pull/910

PR #889, opened 2016-06-25
cleanup i18n files

luclu:
As proposed in https://github.com/ethereum/mist/issues/888.


frozeman:
I don't understand that PR, why are you removing kr?

luclu:
`kr` isn't ISO-639-1. https://github.com/TAPevents/tap-i18n/blob/master/lib/plugin/etc/language_names.js#L64
We already switched to the correct `ko`.

frozeman:
why you added these other languages then?

can you please rebase?

luclu:
Have a look at this table: https://github.com/ethereum/mist/issues/888#issuecomment-228544829
`al` is actually `sq`.

> why you added these other languages then?

Do you mean added to the `project-tap.i18n`?

frozeman:
yes

luclu:
Why not?

frozeman:
Do we have translated language files for those?

luclu:
Yes, for Mist - although Albanian is not complete..

But not for the Wallet.

frozeman:
can you rebasE?

alexvandesande:
Superseded by https://github.com/ethereum/mist/pull/910

PR #813, opened 2016-06-14
fixed #553, adding setting to change UI language

ChihChengLiang:
Dear developers,

Thanks for the awesome project. To address #553 , I added a menu item "settings" that users can change languages in the wallet screen easily.
Note that the language of the app menu can't be changed with this feature, at the moment I can't figure out yet. Hints would be appreciated.

Warm regards,
Chih-Cheng

-------------
update:
successfully changed the language of the app menu.

PR #896, opened 2016-06-27
adding setting to change UI language (fixes #813)

alexvandesande:


ChihChengLiang:
Ahhhh, I'm too late. Just want to point out two things:
1. Adding language names in every `i18n.json` files might not be a good idea. Suppose a new language needs to be added. Either the translation contributor needs to add that language names in every `i18n.json` files, or the dev team have to by somehow inform the contributor to do so and to make sure the job done right. That might be an overhead to the dev team.
Ideally, load the language names from `project-tap.i18n`.

2. The `.substr(0,2)` might fail the `zh-TW` and any further 5-bytes-coded languages everywhere.

ChihChengLiang:
I should move these comments to issues

alexvandesande:
1. Actually that will not be necessary. When not present, the string defaults to english so if there's a language name in english (and the language name is on the native language) then we only need it on the english version

2. Yes, but this is a larger issue. Checking a imperfect match **only** if it doesn't have a better match is complicated

PR #903, opened 2016-06-28
Added NPM shrinkwrap

hiddentao:
This ensures all dependency and sub-dependency versions are fixed.

### DIFF
--- a/npm-shrinkwrap.json
+++ b/npm-shrinkwrap.json
@@ -1,0 +1,2004 @@
+{
+  "name": "Mist",
+  "version": "0.7.6",
+  "dependencies": {
+    "abbrev": {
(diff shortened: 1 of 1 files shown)


frozeman:
How would that work? What does `npm update` do now? is simply `npm install` enough?

luclu:
From my understanding this pretty much omits the `npm update` functionality leaving us with `npm install`.

frozeman:
Yeah, but how would we update?

luclu:
`npm install --save` I guess. @hiddentao?

hiddentao:
Yeah npm install is enough now. But it won't update to the lates versions 
of the packages.
---
Director
HiddenTao Ltd

www.hiddentao.com

PR #905, opened 2016-06-28
fixed webview tagd and updated electrong to 1.2.5

frozeman:
This updates electron to 1.2.5 and also fixes the web view issue, where i had to enable `nodeIntegration` while also removing the require and module object in the preloaded

### DIFF
--- a/README.md
+++ b/README.md
@@ -30,13 +30,13 @@ Once a Mist version is released the Meteor frontend part is bundled using `meteo
 
 Requirements: 
 
-* Electron v1.2.2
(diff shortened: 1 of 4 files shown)


luclu:
Works for me.
What are the downsides of `nodeIntegration: true`? Is this feasible from a security point of view (XSS,..)?

frozeman:
The downside is that you could require any modules, but because i null'ed that its not an issue.

alexvandesande:
Been testing since yesterday without any issues.

PR #914, opened 2016-06-30
fixes bug in which browser had two webviews

alexvandesande:
Many bugs were related to this one, including the browser not detecting accounts and two pages appearing one over the other when no background color was set


alexvandesande:
Superseded by this one: https://github.com/ethereum/mist/pull/911

frozeman:
I would close it, as i already fixed that in #911 which should be merged

Issue #917, opened 2016-07-01
Socket creates multiple event listeners on reset?


frozeman:
Mist version: 0.7.6

This issue is a collection of issues i encountered while working on web3.js subscriptions:

I work currently on the new web3.js subscriptions and especially here on the part which auto-subscribes to a lost socket connection.

It seems like when you have mist running and you kill the node and restart it, that the error/end/timeout etc events get replicated on the socket and the send down multiple times to `owner.send('ipcProvider-....')`.

-----

Additionally there is a uncaught connection error, which leads to mist crashing, when mist is killed.
```js
[2016-07-01 16:37:40.922] [ERROR] main - UNCAUGHT EXCEPTION { Error: connect ENOENT /Users/frozeman/Library/Ethereum/geth.ipc
    at Object.exports._errnoException (util.js:949:11)
    at exports._exceptionWithHostPort (util.js:972:20)
    at PipeConnectWrap.afterConnect [as oncomplete] (net.js:1080:14)
  code: 'ENOENT',
  errno: 'ENOENT',
  syscall: 'connect',
  address: '/Users/frozeman/Library/Ethereum/geth.ipc' }
```

---

- [ ] Multiple event listeners on socket
- [ ] uncaught connection error

hiddentao:
So you're running Geth separately to Mist? And thus shutting it down externally?

hiddentao:
Having trouble reproducing these errors

PR #918, opened 2016-07-01
Added connect event to the IPC backend

frozeman:
This PR, adds the `connect` event to be send down from the IPC backend and also fixes small things in the IPC wrapper.

Ideally we commit fixes for #917 into this branch @hiddentao


hiddentao:
How can I test this out?

frozeman:
Simply add the `web3.currentProvider.on('connect', function)` function and then quit the running node, and restart it after.

hiddentao:
How are you quitting the running node and restarting it? Through Mist? So are you changing network or something?

frozeman:
Canceling manually. If you don't have any idea how to fix the uncaught connection error for now, i would suggest we merge that now and fix the other later

hiddentao:
Yep, agree with doing the merge now.

PR #924, opened 2016-07-05
adds the wallet as the default tab

alexvandesande:


hiddentao:
Why can't I reset the `Tab`s and see these? Some bug.

alexvandesande:
What do you mean?

hiddentao:
So I don't see these new tabs showing. And I think it's because I already have tabs present. I added the `--reset-tabs` CLI option a while ago to reset the list of tabs but that seems to have no effect as far as I can tell. I need to look into it.

Issue #955, opened 2016-07-11
Display human readable function signature for transactions when possible.


pipermerriam:
### What is the problem

Currently, when sending a transaction the raw bytecode of the transaction data is shown in the data field.

<img width="446" alt="screen shot 2016-07-11 at 1 33 28 pm" src="https://cloud.githubusercontent.com/assets/824194/16744017/493988b8-476c-11e6-9dfb-1d77c0476cf7.png">

While showing the bytecode is better than showing nothing, the bytecode is not very helpful to most users and does not expose much useful data in it's raw form.

Ideally, if the transaction that is being sent is calling a function on a contract, the human readable call signature of that function should be shown.
### How could this be fixed.

I suggest querying the human readable version of the 4-byte function signature using the following approach.
1. If the transaction is being executed against a contract who's ABI is known, first look through the function signatures for that contract and see if the first 4 bytes of transaction data match the 4byte signature for any of the contract functions declared within the contract ABI.
2. Next look in a local hard-coded mapping of _common_ function signatures (to be determined what constitutes common).  Current db of ~2000 signatures is 83kb as a json file. (https://gist.github.com/pipermerriam/79429c2163969ff732d5f2aabaaaf0e5)
3. Next query the API of https://www.4byte.directory/ to see if the signature is present in that database (see notes below about this).
### Displaying the function signature

When a human readable signature can be found, I propose displaying it in one of the following ways within the "Execute Contract" modal.
1. In the main title text "Execute Contract" add a sub-title below it that shows the human readable function signature.
2. In the display at the top of the page that shows the two contract icons with the arrow between them and the 4-byte functions signature below them, add the human readable version below the 4-byte function signature.
3. Change the text _"You are about to execute a function on a contract.  This might involve transfer of value"_ to _"You are about to execute the function `transfer(address,uint256)` on a contract.  This might involve the transfer of value"_
4. In the _Data_ section, change the header to be _Data `transfer(address,uint256)`_.
### Thoughts on querying the `4byte.directory` API.

This is a project that I have authored myself.  I'm not in a position to determine whether it is acceptable to add a query to this API, but I will add the following information to hopefully help people make an informed decision.
- The API is hosted on heroku and has simplistic infrastructure.  I expect it to be reliable and stable for the foreseeable future.
- I have no plans to modify the API in a breaking manner.  Any breaking changes will be implemented under a new version namespace.
[...]

PR #962, opened 2016-07-12
Put human readable function signatures in the transaction confirmation screen.

pipermerriam:
https://github.com/ethereum/mist/issues/955

### What was wrong

When sending a transaction, the user will be better informed if they are able to see more data about the transaction.

### How was this improved

Now the function signature is queried against the [`4byte.directory`](https://www.4byte.directory/) api to try and lookup the human readable signature.

#### Cute animal picture

baby-monkeys-in-clothes-wallpaper-3

### DIFF
--- a/interface/client/templates/popupWindows/sendTransactionConfirmation.html
+++ b/interface/client/templates/popupWindows/sendTransactionConfirmation.html
@@ -27,8 +27,8 @@
                     <div class="amount">
                         {{{totalAmount}}} <span class="unit">ETHER</span>
 
-                        {{#if executionFunction}}
-                            <span class="execute-function">{{executionFunction}}</span>
+                        {{#if TemplateVar.get "executionFunction" }}
+                            <span class="execute-function">{{TemplateVar.get "executionFunction"}}</span>
                         {{/if}}
                     </div>
                     <div>
--- a/interface/client/templates/popupWindows/sendTransactionConfirmation.js
+++ b/interface/client/templates/popupWindows/sendTransactionConfirmation.js
@@ -20,6 +20,45 @@ The sendTransaction confirmation popup window template
 @constructor
 */
 
+
+/**
+Takes a 4-byte function signature and does a best-effort conversion to a
+human readable text signature.
+
+@method (lookupFunctionSignature)
+*/
+var lookupFunctionSignature = function(data) {
+    return new Q((resolve, reject) => {
+        if(data && data.length > 8) {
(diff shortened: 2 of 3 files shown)


pipermerriam:
screen shot 2016-07-12 at 2 48 13 pm

pipermerriam:
I would ❤️ feedback on this.  I'm utterly new to this codebase and would like to do things "the right way".  Also won't be offended if someone more familiar wants to take this across the finish line at whatever point they feel that is appropriate.

frozeman:
Im at a conference the next two days, but would surely take a look at the code.. Can you please wait until friday?

frozeman:
Code looks good from the first sight, would need to test it though.
We also need to fix the text size in the image, as it is overlapping

frozeman:
Ideally it would also show not only the function, but the actual values inside the function.

Maybe its better to put this then above the bytes data field e.g.

`transferFrom('0x23456789...', 2000)`

hiddentao:
Missing a label.

luclu:
Ah, my mouse acted weird - having its own will lately.. sorry

evertonfraga:
I am moving forward with this PR, and @pipermerriam  I have a question: Is there a way to get updated versions of the whole dictionary, just like [your Gist](https://gist.github.com/pipermerriam/79429c2163969ff732d5f2aabaaaf0e5)? 

I'm thinking of downloading it on every Mist build.

pipermerriam:
The easiest thing might be for me to provide you with a script that scrapes
the data from the API into your desired format.

pipermerriam:
@evertonfraga Here is a python script that demonstrates scraping the API for all signatures.  I imagine it should be easy to transform into something `node` based.

```
import requests
import json

URL = "https://www.4byte.directory/api/v1/signatures/"


def get_all_signatures():
    page = 1
    while True:
        response = requests.get(URL, {'page': page}, headers={'Content-Type': 'application/json'})
        try:
            response.raise_for_status()
        except requests.exceptions.HTTPError:
            break
        response_data = response.json()
        if response_data['count'] == 0:
            break
        for result in response_data['results']:
            yield result
        page += 1


def write_all_signatures():
    with open('signatures.json', 'w') as f:
        signatures = list(get_all_signatures())
        json.dump(signatures, f)


if __name__ == '__main__':
    write_all_signatures()
```

alexvandesande:
replaced by https://github.com/ethereum/mist/pull/1141

PR #975, opened 2016-07-16
disable eth-node switch until hardfork is supported

luclu:
As @bobsummerwill requested - in case eth won't be HF capable until the HF release - this PR disables the menu-entry.
Though eth `1.2.9` will still be bundled.


bobsummerwill:
Thanks, guys!

PR #976, opened 2016-07-16
update geth 1.4.7 -> 1.4.10

luclu:
@frozeman @karalabe the **linux64** from github won't extract using the current gulp-setup. 
`1.4.7` works like a charm as well as the **linux64** archive from bintray.

There is no error message.
I tested this on OSX as well as Ubuntu 14.04.
I also tried to resemble this on node-REP:
```javascript
var file = 'geth-Linux64-20160716160600-1.4.10-5f55d95.tar.bz2';
streams.push(gulp.src(file).pipe(decompress({strip: 1})).pipe(gulp.dest("./test")));
```
This works fine for all other builds. Omiting `strip: 1` unsurprisingly won't help either.
No bug reports in the **decompress** repos.

Any ideas?

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -40,13 +40,13 @@ var filenameLowercase = 'mist';
 var filenameUppercase = 'Mist';
 var applicationName = 'Mist'; 
 var electronVersion = '1.2.5';
-var gethVersion = '1.4.7';
+var gethVersion = '1.4.10';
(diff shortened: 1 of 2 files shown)


frozeman:
hm, didn't test yet, but i added this already in the DAO-fork PR.

Will try now if it extracts

Issue #987, opened 2016-07-18
remove postinstall hook from package.json


5chdn:
The package.json contains the following:

    "scripts": {
      "ci": "gulp --platform=linux",
      "postinstall": "gulp update-nodes"
    },

Which causes each `npm install` to trigger a `gulp update-nodes` after the installation. Took me a while to figure out where that came from.

In the docs you describe the installation as follows:

> ### Installation

> Now you're ready to install Mist:

> $ git clone https://github.com/ethereum/mist.git
> $ cd mist
> $ git submodule update --init
> $ npm install
> $ gulp update-nodes

Which describes the intentions of `npm install` and `gulp update-notes` to be separated commands. As for now it is not possible to install mist without triggering a full download of all nodes for all operating systems (this is yet another issue).

    [14:03:42] Starting 'clean:nodes'...
    [14:03:42] Finished 'clean:nodes' after 10 ms
    [14:03:42] Starting 'downloadNodes'...
    [14:03:42] Downloading https://github.com/ethereum/go-ethereum/releases/download/v1.4.10/geth-OSX-20160716155225-1.4.10-5f55d95.zip...
    [14:03:42] Downloading https://github.com/ethereum/go-ethereum/releases/download/v1.4.10/geth-Linux64-20160716160600-1.4.10-5f55d95.tar.bz2...
    [14:03:42] Downloading https://github.com/ethereum/go-ethereum/releases/download/v1.4.10/Geth-Win64-20160716155900-1.4.10-5f55d95.zip...
    [14:03:42] Downloading https://bintray.com/karalabe/ethereum/download_file?file_path=geth-1.4.10-stable-5f55d95-linux-386.tar.bz2...
    [14:03:42] Downloading https://bintray.com/karalabe/ethereum/download_file?file_path=geth-1.4.10-stable-5f55d95-windows-4.0-386.exe.zip...
    ^C

Please remove the postinstall hook from `package.json` or provide a convenient way of disabling the `update-nodes` command. Also I would strongly recommend to only download the target OS version of the node.

luclu:
@5chdn could you please try out this PR?

hiddentao:
PR is #988, thanks @luclu

5chdn:
Jeez, what a mess, now I run into the following.

     ERROR: YOUR NODES NEED TO BE UPDATED run $ gulp update-nodes

I do not want to use the mist nodes but the system wide nodes. Please, help a desperate package maintainer :-)

[...]

PR #988, opened 2016-07-18
remove 'update-nodes' hook from 'npm install'; add '--platform' option

luclu:
fixes https://github.com/ethereum/mist/issues/987


luclu:
@hiddentao do we need `npm install -g electron-prebuilt@1.2.2` [.travis.yml#L18](https://github.com/ethereum/mist/blob/develop/.travis.yml#L18) as this seems to get pulled by electron-packager while bundling.

hiddentao:
Yeah we might not need that. I was just copying over our README.md instructions.

Issue #448, opened 2016-03-31
Asking for a password during onboarding seems strange


karalabe:
When I started up a pristine wallet, it asked me for a password that "I should protect like my house". This makes me think that this will be a master password for accessing the wallet. Of course this is actually only the password for an auto generated primary account. However this in very non obvious to someone who didn't actually write Mist.

I think that if Mist detects that there are no accounts present yet, it should offer the user to create one, making it clear that it's inserting a new default account. Further it should be clear that the password is for that new account. Lastly when the account is created, it should display a message that "Yo man: **this** is the account we've just created for you". Because in my experience, I had to wait two hours for Mist to sync up and whatnot to even see what I used my password on.

frozeman:
The on boarding is offering to create a new account, and it shows you right away the account number and identicon.

PR #989, opened 2016-07-18
update 'protectDescription' string to mention etherbase account

luclu:
fix https://github.com/ethereum/mist/issues/448


frozeman:
? There is no main account which pays all the gas.
The gas will be payed by the owner account, or the account your select when creating a wallet.

I would not change the text here, it is fairly clear. 
Etherbase, just means it will be used for mining, which mist is not really build for yet (no mining dapp)

luclu:
Ah, I was always wondering about this design decision. ;P

PR #993, opened 2016-07-19
Fix Mongo sync (LokiDB) race condition

hiddentao:
This fixes the race condition sometimes seen with Mongo sync (i.e. using LokiDB at the backend).


alexvandesande:
Merge included on this: https://github.com/ethereum/mist/pull/1029

PR #1008, opened 2016-07-21
fix compile error always defaults to 'e.message.replace is not a function'

### LINKED ISSUE #1007: Compiler Error always defaults to 'e.message.replace is not a function'
Node type: geth
OS: windows
Network type: main
Mist version: 0.8.1

I try to deploy a contract, but when insert it to the contract deploy window in ethereum wallet (or mist) I get **e.message.replace is not a function** error. The problem here are parts referring to **amIOnTheFork** variable (I have no idea why). When I comment it out error disappears. I also managed to make it compile but the procedure is weird and I guess it shouldn't be doing that.

[Screen Capture](https://drive.google.com/file/d/0B4hSzONfReq8cFBjSXBHSEJiU2M/view?usp=sharing)

```
contract ReplaySafeProxy {
    // Tracks if the contract should work on this chain or not.
    bool enabled = false;
    // Fork oracle to use
    AmIOnTheFork amIOnTheFork = AmIOnTheFork(0x2bd2326c993dfaef84f696526064ff22eba5b362);
    address target;
    // _target is the address this contract will forward all funds to.
    // forked specifies if it should work on ETH (true) or ETHC (false)
    // Any value transfer to this contract will fail on the other network.
    function ReplaySafeProxy (address _target, bool forked) {
        if (amIOnTheFork.forked() == forked) {
            enabled = true;
        }
        target = _target;
    }
    // Forward value transfers.
    function() {
        if (enabled == false || msg.value == 0 || !target.send(msg.value)) {
            throw;
        }
    }
}
```

luclu:
fix #1007

### DIFF
--- a/modules/ipc/ipcProviderBackend.js
+++ b/modules/ipc/ipcProviderBackend.js
@@ -337,7 +337,7 @@ class IpcProviderBackend {
 
             if (e.message) {
                 e.error = {
-                    message: e.message.replace(/'[a-z_]*'/i, "'"+ item.method +"'")
+                    message: e.message
                 };
 
                 delete e.message;


luclu:
regex pattern: `/Error: :([0-9]+:)+ /`
array: `e.message[0].replace` ... more tomorrow ;)

hiddentao:
@luclu I've done the extended fix in #1009 .

PR #1015, opened 2016-07-22
Temporary fix, as per Electron issue #5110

evertonfraga:
When loading a Dapp with transparent appBar, sometimes its ```object``` tag (inside ```webview```) appears cropped, with different heights. It is adjusted back to normal upon window resize.

screenshot 2016-07-22 17 48 22

Digging a bit, and now I can fully reproduce that, it happens when ```ipc.sendToHost('appBar', 'transparent')``` is fired while the referenced Webview is hidden. So Electron can't properly calculate the not visible ```object```'s new size.

I found a temp fix which consists in changing tab's css rules and it seemed to work. I'll follow this issue for future fixes on Electron: https://github.com/electron/electron/issues/5110

### DIFF
--- a/interface/client/styles/layout.import.less
+++ b/interface/client/styles/layout.import.less
@@ -152,11 +152,11 @@ main {
     &.hidden {
         // visibility: hidden;
         z-index: 1;
-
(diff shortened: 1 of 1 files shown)


alexvandesande:

PR #998, opened 2016-07-19
fixed #901 , mist now get best matched language code

ChihChengLiang:
hi~
I fixed #901 by adding a `getBestMatchedLangCode` function in i18n module. The mist now get the closest language code from supported language.

Better test this fix with a computer whose default language is `en-US`.

PR #1018, opened 2016-07-23
Fixed #1000 by removing tap:i18n-bundler

### LINKED ISSUE #1000: Can't switch to German with `view > change language`
Node type: geth
OS: linux
Network type: main/test
Mist version: 0.8.1
branch: develop

Can't switch to German with `view > change language`. At the first glance, the `TAPi18n.setLanguage("de")` is not behave normally.
@luclu

ChihChengLiang:
Removed tap:i18n-bundler and the i18n should work fine.

`tap:i18n-bundler` is a meteor package that bundles all i18n resources to the `public` directory on meteor startup. It aimed to make client preload i18n files and make these files available offline for Cordova. Base on the serverless design principle of Dapp, I believe that feature introduces more troubles than benefits for the mist browser.

The root cause of #1000 is that i18n-bundler failed to update languages to `app/public/i18n`.

Please review this PR with https://github.com/ethereum/meteor-dapp-wallet/pull/250 together.

### DIFF
--- a/interface/.meteor/packages
+++ b/interface/.meteor/packages
@@ -6,7 +6,6 @@
 
 less
 tap:i18n
-tap:i18n-bundler
 jeeeyul:moment-with-langs
 raix:handlebar-helpers
 frozeman:animation-helper
--- a/interface/.meteor/versions
+++ b/interface/.meteor/versions
@@ -91,7 +91,6 @@ standard-minifier-css@1.0.7
 standard-minifier-js@1.0.7
 standard-minifiers@1.0.6
 tap:i18n@1.8.2
-tap:i18n-bundler@0.3.0
 templating@1.1.10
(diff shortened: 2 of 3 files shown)


luclu:
This works great for me! 
Language selection seems to be fixed with those steps:

 - this PR
 - that PR #998
 - deleted the two folder: (they won't get recreated as the bundler is removed)
```bash
interface/public/i18n/
interface/public/public/i18n/
```

alexvandesande:
Not all languages are switching but I think this is more related to incomplete translations than anything else. I'm happy to see we can do away with the bundlers..

luclu:
@alexvandesande yes I notices the same, but those are the languages are not available yet in meteor-wallet-dapp.

alexvandesande:

Issue #1002, opened 2016-07-20
MIST UI v0.8.1 not starting (hangs on meteor starting up..)


FlepNL:
Node type: geth.exe
OS: windows 10 - 64bit
Network type: main
Mist version: 0.8.1

New Mist Wallet UI is not starting up anymore, after being able to start up a couple of times before.
Ethereum Wallet version 0.8.0 is starting up.

In order to start any of the two wallets, I startup geth.exe via the command line.

The console log when executing mist.exe from the command line displays the following:

<img width="664" alt="console output 1" src="https://cloud.githubusercontent.com/assets/20551360/16976547/25f1ff62-4e82-11e6-8320-32037294861d.PNG">
<img width="663" alt="console output 2" src="https://cloud.githubusercontent.com/assets/20551360/16976557/309fac84-4e82-11e6-9406-5328bd2830be.PNG">

From the node.log:
```
I0720 13:07:16.064942 ethdb/database.go:82] Alloted 512MB cache and 1024 file handles to C:\Users\MacFlep\AppData\Roaming\Ethereum\chaindata
I0720 13:07:18.805790 cmd/utils/flags.go:842] ------------------------------------------------------------------------------------------------
I0720 13:07:18.805790 cmd/utils/flags.go:843] Geth is currently configured to SUPPORT the DAO hard-fork!
I0720 13:07:18.805790 cmd/utils/flags.go:844] You can change your choice prior to block #1920000 with --support-dao-fork or --oppose-dao-fork.
I0720 13:07:18.805790 cmd/utils/flags.go:845] After the hard-fork block #1920000 passed, changing chains requires a resync from scratch!
I0720 13:07:18.805790 cmd/utils/flags.go:846] ------------------------------------------------------------------------------------------------
I0720 13:07:18.811797 ethdb/database.go:169] closed db:C:\Users\MacFlep\AppData\Roaming\Ethereum\chaindata
I0720 13:07:18.824820 ethdb/database.go:82] Alloted 512MB cache and 1024 file handles to C:\Users\MacFlep\AppData\Roaming\Ethereum\chaindata
I0720 13:07:18.884905 ethdb/database.go:82] Alloted 16MB cache and 16 file handles to C:\Users\MacFlep\AppData\Roaming\Ethereum\dapp
I0720 13:07:18.895941 eth/backend.go:172] Protocol Versions: [63 62], Network Id: 1
I0720 13:07:18.897959 eth/backend.go:201] Blockchain DB Version: 3
[...]

PR #1026, opened 2016-07-24
Fixes Meteor startup error having a redirect to about:blank

lks128:
Similarly like in the issue #1002 I had an error while starting Mist, and it turned out that the problem was a strange redirect to the about:blank page (according to the code I've seen, .redirect should be equal to the .url). Interestingly, if the redirect is something like https://google.com, nothing bad happens.

```
{ position: 0,
  url: 'https://ethereum.org/',
  _id: 'browser',
  webviewId: 14,
  icon: null,
  appBar: null,
  name: 'Ethereum Project',
  permissions: { accounts: [ '0xccca739ae27bfaef223cbe58b665860d691de087' ] },
  redirect: 'http://about:blank',
  menuVisible: true }
```

This PR adds a quick fix on minimongo frontendSync, which removes a redirect to the about:blank page. (don't know if it has to remove it always or only if url != redirect)

I agree that the root of the problem is deeper, but I think that in general it makes the application more stable.

Let me know what do you think.

### DIFF
--- a/modules/syncMinimongo.js
+++ b/modules/syncMinimongo.js
@@ -114,6 +114,14 @@ exports.frontendSync = function(coll) {
         coll.remove({});
 
         JSON.parse(dataStr).forEach(function(record) {
+            // On Meteor startup if a record contains a redirect to about:blank
+            // page, the application process crashes.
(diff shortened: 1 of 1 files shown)


alexvandesande:

Issue #828, opened 2016-06-15
Make new window loader screen frameless


alexvandesande:
Remove the buttons from the loading new window  screen.

<img width="172" alt="captura de tela 2016-06-15 as 7 12 13 pm" src="https://cloud.githubusercontent.com/assets/112898/16099176/32d6e74e-332d-11e6-8ed8-20cefbb1ecb4.png">

hiddentao:
I'll fix this for 0.8.2

luclu:
I already digged into this a little.
The problem seems to be upstream with the program flow in [native_window_mac.mm](https://github.com/electron/electron/blob/master/atom/browser/native_window_mac.mm) - as many PRs regarding window decoration actually are.

It is possible to circumvent it in Mist by commenting:
 - https://github.com/ethereum/mist/blob/f4dbbaf8cb0008f621ff0e54c00b00e99e26982c/modules/windows.js#L37
 - https://github.com/ethereum/mist/blob/f4dbbaf8cb0008f621ff0e54c00b00e99e26982c/modules/windows.js#L186

Though this will break the titlebars across OSs.

hiddentao:
Thanks @luclu. I think we can fix by setting `titleBarStyle` to empty string just for the loading window.

hiddentao:
Fixed in #1033

luclu:
>  I think we can fix by setting titleBarStyle to empty string just for the loading window.

No, that is just too simple and obvious. Please try harder on that one.

PR #1033, opened 2016-07-25
Fix for #828

hiddentao:
As the title says.


luclu:
Tschush...

luclu:
Works on all platforms.

PR #1047, opened 2016-07-28
closes #264, check for system wide installation of geth

### LINKED ISSUE #264: Look for a system wide installation of geth or eth

luclu:
This current state of this PR will make `getNodePaths.js - getNodePaths()` to 
 - check for nodes in `./nodes/<type>/<platform>/<type>`
 - check if system-installation of that node is available (Linux + OS X)
 - gather and compares their versions
 - stores the relevant (latest version's) paths in a `resolvedPaths[type]` dictionary

Ideally this should be run early during startup as it can take up to 2 seconds to resolve the versions for (geth, eth and parity installed).


gulpfile.js
```diff
@@ -40,7 +40,7 @@ var filenameLowercase = 'mist';
 var filenameUppercase = 'Mist';
 var applicationName = 'Mist'; 
 var electronVersion = '1.2.5';
-var gethVersion = '1.4.10';
+module.exports.gethVersion = '1.4.10';
```
luclu: This feels ugly + travis doesn't like this:
```
/home/travis/build/ethereum/mist/gulpfile.js:213

        if(!!~file.path.indexOf('-'+ gethVersion +'-')) {

                                     ^
```

modules/getNodePath.js
```diff
@@ -25,15 +27,44 @@ module.exports = function(type) {
...
+                
+                var systemGethVersion;
+                setTimeout(function () {
+                    if(systemGethPath !== null) {
+                        log.error(systemGethPath);
+                        var proc = exec(systemGethPath + ' version', 
+                            function(error, stdout, stderr) {
+                                systemGethVersion = stdout.match(/(?:[\d.]+)/)[0];
+                                log.warn('Geth node version: ' + systemGethVersion);
+                                log.warn(gulpfile.gethVersion);
```
luclu: What is a better way to pull the internal version of geth?
 - creating an additional local variable?
 - pulling the version with `geth version`, too?

modules/getNodePath.js
```diff
@@ -25,15 +27,44 @@ module.exports = function(type) {
...
+                var proc = exec('which geth', 
+                    function(error, stdout, stderr) {
+                        systemGethPath = stdout.toString().replace(/\n/, '');
+                        if (error !== null) {
+                                log.error("error exec 'which geth': " + error + stdout);
+                        }
+                    });
+                
+                var systemGethVersion;
+                setTimeout(function () {
```
luclu: I currently use `child_process.exec` to execute the two commands `which geth` and `geth version` and use a timeout to wait for the async call to finish. 
What better approaches are there? Especially considering the later program flow depends on its output?

gulpfile.js
hiddentao: My fix for the Linux node download issue will replace this with something better.

modules/getNodePath.js
```diff
@@ -25,15 +27,44 @@ module.exports = function(type) {
...
+
     if (globallySetType) {
         resolvedPaths[type] = globallySetType;
     } else {
         var binPath = (Settings.inProductionMode)
             ? binaryPath.replace('nodes','node') + '/'+ type +'/'+ type
             : binaryPath + '/'+ type +'/'+ process.platform +'-'+ process.arch + '/'+ type;
 
-        if(Settings.inProductionMode) {
+        if(/*Settings.inProductionMode*/ true) {
```
hiddentao: ?? Perhaps we need to get rid of the condition altogether then?

modules/getNodePath.js
hiddentao: I would recommend against using `which`. The `type` command is probably what we need, see http://superuser.com/questions/351889/what-is-the-unix-command-to-find-out-what-executable-file-corresponds-to-a-given

modules/getNodePath.js
hiddentao: Using `geth version` for the internal geth exec sounds good to me. This is also a way of verifying that it executes properly on the local OS.

modules/getNodePath.js
luclu: Yeah this is just for testing.
Sounds good to me.

modules/getNodePath.js
luclu: Ah great, I actually run into the [non-existing exit code issue](https://stackoverflow.com/questions/592620/check-if-a-program-exists-from-a-bash-script).

luclu:
This current state of this PR will make `getNodePaths.js - getNodePaths()` to 
 - check for nodes in `./nodes/<type>/<platform>/<type>`
 - check if system-installation of that node is available
 - gather and compares their versions
 - stores the relevant (latest version's) paths in a `resolvedPaths[type]` dictionary

Ideally this should be run early during startup as it can take up to 2 seconds to resolve the versions for (geth, eth and parity installed).

luclu:
argh...

PR #1060, opened 2016-08-01
Remove fork question, and add legacy chain warning

frozeman:
This PR, will remove the chain selection question and add a info for people on the legacy chain, on how to upgrade, or refers them to the ethereum classic community.


bezanes12:
Hi
Sory to disturb youToday i have buy 100euros ETC on bilboat.net with paysafecard.

But i havn't receive ETC on MyEtherWallet. The 100 euros are deducted 
from my bank account and bilboat tell me that the ETC is sent.

Someone has the same problem?
Is there any settings or other to do on ETC because it's new?

interface/i18n/mist.en.i18n.json
```diff
@@ -83,6 +83,10 @@
                 "win32": "Go to your Internet Time Settings in your system preferences and check the box. See this guide for details: http://www.guidingtech.com/3119/windows-clock-sync/",
                 "linux": "To enable a time sync server install \"ntp\" via \"apt-get install ntp\".",
                 "darwin": "To enable time sync, open the time preferences and check \"Set the time and date automatically\"."
+            },
+            "legacyChain": {
+                "title": "Legacy chain detected",
+                "description": "Your node is currently on the unsupported ethereum classic chain. To use this chain use tools provided by the ethereum classic project at\nhttp://ethereumclassic.github.io.\n\nTo revert to the main ethereum chain follow the tutorial here:\nhttp://github.com/ethereum/mist/releases/0.8.2"
```
hiddentao: * Please capitalize, i.e. "Ethereum Classic"
* Use HTTPS: https://ethereumclassic.github.io/
* _To revert...[please] follow the tutorial..._
* Use HTTPS: https://github.com/ethereum/mist/releases/0.8.2

modules/ethereumNode.js
```diff
@@ -527,7 +517,7 @@ class EthereumNode extends EventEmitter {
         try {
             return fs.readFileSync(fullPath, {encoding: 'utf8'});
         } catch (err){
-            log.error(`Unable to read from ${fullPath}`, err);
+            log.error(`Unable to read from ${fullPath}`);
```
hiddentao: Why has `err` been removed?

frozeman: This error will always no appear for installations where it can't load the `daoFork` file. To make this error less huge in the console i removed the `err` object, especially if the loading file error should be pretty clear why it happens.

Otherwise we could build in a check, if `daoFork` don't log the error

hiddentao: It would be good to see an error if it's due to file being unreadable, etc. What we can do is do a pre-check using `fs.accessSync(fullPath, fs.R_OK)` - if this throws then it means the file can't be read. If it passes then we go ahead and do `fs.readFileSync`. If that fails we output the error.

frozeman:
I added that it opens the rebase page with instructions, once the user clicks OK, and it closes the app

modules/ethereumNode.js
```diff
@@ -281,12 +281,10 @@ class EthereumNode extends EventEmitter {
                 this._saveUserData('node', this._type);
                 this._saveUserData('network', this._network);
 
-                // FORK RELATED
-                this._saveUserData('daoFork', this.daoFork);
 
-                return this._socket.connect(Settings.rpcConnectConfig, {
-                    timeout: 30000 /* 30s */
-                })  
+                return this._socket.connect({ path: Settings.rpcIpcPath}, {
```
hiddentao: Why has this changed? `Settings.rpcConnectConfig` is the correct one.

modules/ethereumNode.js
```diff
@@ -513,6 +504,13 @@ class EthereumNode extends EventEmitter {
 
         log.trace('Load user data', fullPath);
 
+        // check if the file exists
+        try {
+            fs.accessSync(fullPath, fs.R_OK);
+        } catch (err){
+            return;
```
hiddentao: Better if this returns `null`, same as the `return` statement further down.

alexvandesande:
Ready to be merged In my opinion. Will do tomorrow morning unless there's any opposition

hiddentao:
Taking a closer look at this now. Some things I want to change.

* The check for daoFork = false should take place before initialising the connection to the node
* The code calls `ethereumNode._loadUserData` from outside `ethereumNode`. Methods prefixed with `_` are meant to be internal and so I want to refactor this call and these methods so that they can be publicly accessible.

hiddentao:
I've commited fixes for the above.

mintzhao:
*F word!!!!*

Where can i know how to use 0.8.2? [The URL in the app](https://github.com/ethereum/mist/releases/0.8.2)   Can you guys find out a tutorial to teach how to revert to the main ethereum chain???????????????????????? 

just go back and using etc anyway, it seems etc is more popular

evertonfraga:
Hi @mintzhao. By the end of [this page](https://github.com/ethereum/mist/releases/0.8.2) you can find Mist 0.8.2 to use with Ethereum (ETH) chain the majority of the community has chosen to follow.

If by any chance you want to use Mist on a ETC chain, please go to their Mist repo (hint: not this repository) and open an issue there, as they have full support for ETC chain. Worth noticing ETC is not backed by the Ethereum Foundation, although the name might suggest it is.

And lastly, I'd ask you to please behave yourself. This is a public github, keep your bad words to a proper place. Ok? 

Cheers.

mintzhao:
ok, my bad. sorry,I will delete it.

PR #972, opened 2016-07-15
Integrating electron-builder for better distributables building

hiddentao:
This replaces `electron-packager` with [electron-builder](https://github.com/electron-userland/electron-builder) so that we can build distributables with installers for both Mist and the Wallet.

* Creates installers for Windows, Mac and Linux (debian) 👊 
* Builds for 32-bit on all platforms where supported
* Streamlined node download and unzipping process
* Renames `darwin` to `mac` inside code and gulpfile because that's what electron-builder uses
* Upon success, builds are placed into either `dist_mist/release` or `dist_wallet/release` depending on chosen app type (mist, wallet)

PR #1064, opened 2016-08-03
check for system-wide installation of nodes; dynamically populate node-menu

### LINKED ISSUE #264: Look for a system wide installation of geth or eth

luclu:
closes #264

This current state of this PR will make `getNodePaths.js` to:
- `getNodePaths.probe()`
  - enumerate bundled nodes
  - check if system-installation of the relevant nodes are available (only Linux and mac)
  - gather and compare their versions
  - store the relevant (latest version's) paths in a `resolvedPaths[type] => {type:{path:version}}` dictionary
  - takes about 800ms with bundled + system instances of geth + eth (macbook pro, SSD)
- `getNodePaths.query(nodeType)`
  - returns preferred path of node

Prefers the latest version of a chosen node. Also system installation of node has higher priority. 
#### TODO
- [x] abstract to make drag and drop addition of any node possible (like parity)
- [x] make develop and production compatible
- [x] use promises
  - [x] to collect external command's outputs
    - [ ] refactor promise structure (race conditions)
  - [x] to integrate with `ethereumNode.js`
- [x] debug log entries
- [x] start `probe()` as soon as initial node connection times out (don't wait the 3 seconds)
  - [x] promisify to prevent timeout
- [x] basic error handling
- [x] populate node menu labels dynamically
  - [x] add version string
  - [x] hide if only one node type is available 
  - [ ] initiate rebuild of appMenu (gives an error); don't wait for node startup to trigger repopulate
- [ ] rebase on #972 and #1228 once merged

### DIFF
--- a/interface/i18n/mist.en.i18n.json
+++ b/interface/i18n/mist.en.i18n.json
@@ -59,6 +59,8 @@
                 "runTests": "Run tests",
                 "logFiles": "Show log file",
                 "ethereumNode": "Ethereum Node",
+                "starting": "starting",
+                "externalNode": "using external node",
                 "network": "Network",
                 "mainNetwork": "Main Network",
                 "startMining": "⛏ Start Mining (Testnet only)",
--- a/modules/ethereumNode.js
+++ b/modules/ethereumNode.js
@@ -41,8 +41,6 @@ class EthereumNode extends EventEmitter {
         this._type = null;
         this._network = null;
 
-        getNodePath();
-
         this._socket = Sockets.get('node-ipc', Settings.rpcMode);
 
         this.on('data', _.bind(this._logNodeData, this));
@@ -329,13 +327,15 @@ class EthereumNode extends EventEmitter {
         this._network = network;
         this._type = nodeType;
 
-        const binPath = getNodePath(nodeType);
-
-        log.debug(`Start node using ${binPath}`);
-
         return new Q((resolve, reject) => {
-            this.__startProcess(nodeType, network, binPath)
-                .then(resolve, reject);
+            getNodePath.probe()
+            .then(() => {
+                const binPath = getNodePath.query(nodeType);
+                log.debug(`Start node using ${binPath}`);
+
+                this.__startProcess(nodeType, network, binPath)
+                    .then(resolve, reject);
+            });
         });
     }
 
--- a/modules/ethereumNode.js
+++ b/modules/ethereumNode.js
@@ -330,7 +330,7 @@ class EthereumNode extends EventEmitter {
         return new Q((resolve, reject) => {
             getNodePath.probe()
             .then(() => {
-                const binPath = getNodePath.query()[nodeType];
+                const binPath = Object.keys(getNodePath.query()[nodeType])[0];
                 log.debug(`Start node using ${binPath}`);
 
                 this.__startProcess(nodeType, network, binPath)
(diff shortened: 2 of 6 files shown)


alexvandesande:
How's the progress here Luca? You seem to be doing some interesting stuff

5chdn:
Tested on ArchLinux. :+1:

evertonfraga:
@luclu any status on this PR? i can test along with #1228 if needed.

luclu:
@hiddentao what functionality this PR provides is not available via the new node-manager and thus should be added to it if desirable?

hiddentao:
@luclu The drag-and-drop support for Parity - how exactly will that work?

The new client-binaries library would be able to scan for download Parity binaries. But at present we have no UI for letting the user choose what client they wish to use. I was wondering if adding this is a good idea - e.g. we can provide a command-line flag which tells Mist to first show a window showing available clients for the system (availability calculated using client-binaries). The user would then be able to choose which client they want to use for that run. Of course, one issue with doing this is that since Geth and Parity use different chaindata and keystore folders the user will not be able to seamlessly pick up where they left off with the previous client in terms of chain sync.

What do you think?

luclu:
@hiddentao I think all functionality of this PR was superseded  by the current implementation of the clientBinaryManager.

I will close this.

PR #1110, opened 2016-08-24
Fixes blank screen at startup

evertonfraga:
Fixes the race condition at startup, where the `mistInit()` was expected to have the DB synced before insert the Tabs from the database.

It ensures `syncMinimongo.frontendSync` resolves a promise when the collections are empty.

### DIFF
--- a/interface/client/appStart.js
+++ b/interface/client/appStart.js
@@ -43,9 +43,6 @@ mistInit = function(){
                 admin: true
             }
         });        
-    })
-    .then(function() {
-        window.trigger('mist-ready');
     });
 };
 
--- a/interface/client/appStart.js
+++ b/interface/client/appStart.js
@@ -4,11 +4,6 @@
 if(location.hash)
     return;
 
-
-// set browser as default tab
-if(!LocalStore.get('selectedTab'))
-    LocalStore.set('selectedTab', 'wallet');
-
 /**
 The init function of Mist
 
@@ -16,7 +11,6 @@ The init function of Mist
 */
 mistInit = function(){
     console.info('Initialise Mist');
-
 
     Tabs.onceSynced.then(function() {
         if (0 <= location.search.indexOf('reset-tabs')) {
@@ -35,14 +29,20 @@ mistInit = function(){
             });
         }
 
-
         Tabs.upsert({_id: 'wallet'}, {
             url: 'https://wallet.ethereum.org',
             position: 1,
             permissions: {
                 admin: true
             }
-        });        
+        });
+
+        // Sets browser as default tab if:
+        // 1) there's no record of selected tab
+        // 2) data is corrupted (no saved tab matches localstore)
+        if(!LocalStore.get('selectedTab') || !Tabs.findOne(LocalStore.get('selectedTab'))){
+            LocalStore.set('selectedTab', 'wallet');
+        }
     });
 };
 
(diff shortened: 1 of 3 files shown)


hiddentao:
Looks good to me.

evertonfraga:
If a user has some non-default tab selected and wipes their database, Mist would try (and fail) to select that tab, resulting in a blank screen. This PR also fixes this bug (see appStart.js).

modules/db.js
```diff
@@ -16,9 +16,7 @@ exports.init = function() {
         // if db file doesn't exist then create it
         try {
             log.debug(`Check that db exists: ${filePath}`);
-
-            fs.accessSync(filePath, fs.R_OK);
-
+            fs.accessSync(filePath, fs.W_OK);
```
evertonfraga: @hiddentao here i've changed to check if the file is writeable. we could circumvent permission errors more easily.

hiddentao: Yeah, saw that. That looks fine to me. We could do `fs.R_OK | fs.W_OK` I suppose.

evertonfraga: good idea. added to pr.

interface/client/appStart.js
```diff
@@ -35,17 +29,20 @@ mistInit = function(){
...
         Tabs.upsert({_id: 'wallet'}, {
             url: 'https://wallet.ethereum.org',
             position: 1,
             permissions: {
                 admin: true
             }
-        });        
-    })
-    .then(function() {
-        window.trigger('mist-ready');
```
alexvandesande: Yes, this was also bugging me and giving an error. @hiddentao do you remember what it was for?

modules/syncMinimongo.js
```diff
@@ -100,34 +90,30 @@ exports.backendSync = function() {
...
 
 exports.frontendSync = function(coll) {
-    var ipc = electron.ipcRenderer;
+    var ipc = electron.ipcRenderer,
+        collName = coll._name,
+        syncDoneResolver;
 
-    var collName = coll._name;
-
-    console.debug('Reload collection from backend: ', collName);
```
hiddentao: why has this line been removed?

modules/syncMinimongo.js
```diff
@@ -100,34 +90,30 @@ exports.backendSync = function() {
...
     (new Promise(function(resolve, reject) {
-        var dataStr = ipc.sendSync('minimongo-reloadSync', {
+        var dataStr, dataJson;
+
+        dataStr = ipc.sendSync('minimongo-reloadSync', {
             collName: collName
         });
 
-        if (!dataStr) {
+        if (!dataStr || (dataJson = JSON.parse(dataStr)).length == 0) {
```
hiddentao: JSON.parse() needs to go inside the `try` block below where it originally was in case it throws an Error.

interface/client/appStart.js
hiddentao: It was for future usage. A dapp would be able to listen to the `mist-ready` event to know when all data was loaded and ready. I'd prefer if this was re-instated.

evertonfraga:
I'll perform both fixes

interface/client/appStart.js
alexvandesande: but trigger isn't even a valid function. Did you meant `$(window).trigger` ?

alexvandesande:
It saves the collections correctly, but I never had the original problem

PR #1114, opened 2016-08-25
Filter result of eth.accounts even for batch IPC calls

hiddentao:
This fixes a security issue raised earlier. Although `eth.accounts` are filtered according to what's visible to the active Mist tab it was still possible to retrieve all accounts by making the same call as part of a batch IPC request. This PR fixes that.

### DIFF
--- a/modules/ipc/ipcProviderBackend.js
+++ b/modules/ipc/ipcProviderBackend.js
@@ -289,7 +289,10 @@ class IpcProviderBackend {
                         ? this._processors[finalPayload.method]
                         : this._processors.base;
 
-                    return processor.exec(conn, nonErrorPayloads);
+                    return processor.exec(
+                        conn, 
+                        isBatch ? nonErrorPayloads : nonErrorPayloads[0]
+                    );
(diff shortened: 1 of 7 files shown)


modules/ipc/ipcProviderBackend.js
```diff
@@ -26,6 +26,7 @@ const ERRORS = {
     METHOD_TIMEOUT: {"code": -32603, "message": "Request timed out for method  \'__method__\'."},
     TX_DENIED: {"code": -32603, "message": "Transaction denied"},
     BATCH_TX_DENIED: {"code": -32603, "message": "Transactions denied, sendTransaction is not allowed in batch requests."},
+    BATCH_COMPILE_DENIED: {"code": -32603, "message": "Compilation denied, compileSolidity is not allowed in batch requests."},
```
hiddentao: This wasn't necessary for the fix but I added it anyway. You can't do a contract compilation as part of a batch call because we don't support that.

hiddentao:
More fixes added.

Issue #821, opened 2016-06-14
Replace the word "Browser" when discussing Mist


doppio:
Alex Van de Sande has stated that he doesn't want users to think of Mist as a "browser for Ethereum" and to compare Mist with web browsers like Chrome or Firefox. Using the term "browser" on the GitHub page and elsewhere contradicts this notion. I was under the imperssion for a long time that Mist was trying to achieve the same kinds of features as Chrome. This could really confuse less tech-savvy users as the Ethereum community grows.

Some alternative suggestions, possibly prefaced by the word "Dapp":
- Hub
- Explorer
- Directory
- Viewer
- Center

alexvandesande:
Mist Navigator

derektiffany:
Mist Stratus

PR #1117, opened 2016-08-25
Change README.md to use 'Navigator' instead of 'Browser'

doppio:
Update the README to use the term 'Navigator' rather than 'Browser' to avoid confusion with web browsers like Chrome or Firefox. See Alex's comment on #821.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,10 +1,10 @@
-# Mist Browser
+# Mist Navigator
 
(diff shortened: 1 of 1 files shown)


alexvandesande:
I like the idea but I'm not sure there's much public support:

https://twitter.com/avsa/status/756870679086174208

doppio:
Ah, okay. What do you think about just removing the word "Browser" altogether as a couple people suggest on the poll? May be worth running another poll on that?

hiddentao:
I'd probably leave it as Browser too as it's easy for people hearing to understand what it does.

luclu:
I personally have to consent with @hiddentao. I regard Ethereum to be already conceptionally complicated enough as to elide these kind of skeuomorphism.

alexvandesande:
I'm closing this for now. Although I still have a nice feeling towards Navigator!..

PR #1138, opened 2016-08-31
Use recent version of Solidity

chriseth:
The version used was over four months old, several important bugs were fixed in the meantime. I did not test whether this change breaks anything.

### DIFF
--- a/package.json
+++ b/package.json
@@ -23,7 +23,7 @@
     "numeral": "^1.5.3",
     "os-timesync": "^1.0.6",
     "semver": "^5.1.0",
-    "solc": "^0.3.1-1",
+    "solc": "^0.3.6",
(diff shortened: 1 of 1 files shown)


christoph2806:
as far as I can see, the compiler runs without optimization - is that for purpose?

chriseth:
I'm not sure if it is on purpose, but it is at least safer.

luclu:
@chriseth release `0.8.2` is bundled with solc `0.3.5` ([npm-shrinkwrap.json#L2821](https://github.com/ethereum/mist/blob/develop/npm-shrinkwrap.json#L2821). 
We switched to npm-shrinkwrap.json instead of package.json recently to have better control about the modules dependency tree.
Please feel free to update this PR to `0.3.6` wrap-style.

PR #1141, opened 2016-09-01
New confirmation pane

alexvandesande:


alexvandesande:
Ready to be merged

.editorconfig
```diff
@@ -0,0 +1,9 @@
+root = true
```
luclu: The editor config can be only locally excluded by using `.git/info/exclude` instead of `.gitignore` (same syntax).

luclu:
@alexvandesande here is some cropped text-field I ran into:

 bildschirmfoto 2016-09-08 um 22 44 10

luclu:
There seems to be some issue with the SSL access to 4bytes when trying to decode the data (main-net and test-net):

bildschirmfoto 2016-09-08 um 23 34 59

Also at some point in time using this branch I could not input any text in any text-field anymore until I restarted Ethereum-Wallet. I will try to observe this and reproduce.

Issue #1035, opened 2016-07-25
Change fullscreen shortcut to follow native conventions


evertonfraga:
F11 Windows
F11 Ubuntu (afaik)
Ctrl+Cmd+F OSX

#### System information
``````
Version: `0.8.1`
OS & Version: windows/linux/osx
``````

alexvandesande:
@evertonfraga what's the status on this?

evertonfraga:
unstarted

PR #1151, opened 2016-09-08
Fullscreen shortcut

evertonfraga:
Wasn't able to test on my VMs tho. had problems with keystrokes configuration — F11 key needed. 

See #1035.


evertonfraga:
It's working in Linux - tested in Ubuntu 64 VirtualBox.

PR #1155, opened 2016-09-12
Swarm Support

alexvandesande:
This is a stepping stone for swarm support when it’s released. Right now it temporarily masks the localhost URL on the interface, but the intent is to change this to the actual connection behind the scenes


frozeman:
This contains all the changes from the confirmation window PR?

interface/client/appStart.js
```diff
@@ -24,7 +24,7 @@ mistInit = function(){
 
             Tabs.insert({
                 _id: 'browser',
-                url: 'https://ethereum.org',
+                url: 'http://dapps.ethercasts.com/',
```
evertonfraga: @alexvandesande who takes care of dapps.ethercasts? Should we contribute with them to have a nicer (and less colorful) layout? 
Will this be officially the Mist first page?

hiddentao:
New CLI option: `bzzhost`. This points to the Swarm host base URL. Default value is `http://localhost:32200`. If your Swarm is running differently then pass it in using this CLI option.

interface/client/appStart.js
alexvandesande: > Will this be officially the Mist first page?
 
No. I just think it's a better placeholder than ethereum.org. We need to create a proper one. I would prefer that it would be exclusive to .eth or swarm uris.

alexvandesande: Unless there's some sort of legal issue on pointing to third party content, of course. I'll check it with Ming.

modules/customProtocols.js
```diff
@@ -0,0 +1,46 @@
...
+const electron = require('electron');
+const protocol = electron.protocol;
+const path = require('path');
+const Settings = require('./settings');
+const log = require('./utils/logger').create('customProtocols');
+
+
+const protocols = {
+    mist: (req, cb) => {
+        const url = (req.url.indexOf('mist://interface') !== -1)
```
alexvandesande: Interesting. What would be this used for?

hiddentao: This was already there so I kept it. It looks like it's a way for people to browse to specific screens within Mist via a URL.

gulpfile.js
```diff
@@ -522,9 +523,33 @@ gulp.task('getChecksums', [], function(done) {
 });
 
 
+gulp.task('download-signatures', function(cb) {
+    got('https://www.4byte.directory/api/v1/signatures/?page_size=20000&ordering=created_at', {
```
frozeman: Why is this part of the Swarm PR?

Issue #1208, opened 2016-09-26
Node tries to install modules for mac and win on linux.


5chdn:
#### System information
``````
Version: `0.8.3`
OS & Version: linux `4.7.4` (ArchLinux)
``````

Running `npm install` on a linux host, tries to install the 7zip binary packages for mac and win which are not supported:

    npm WARN optional SKIPPING OPTIONAL DEPENDENCY: 7zip-bin-win@^1.0.0 (node_modules/7zip-bin/node_modules/7zip-bin-win):
    npm WARN notsup SKIPPING OPTIONAL DEPENDENCY: Unsupported platform for 7zip-bin-win@1.0.0: wanted {"os":"win32","arch":"any"} (current: {"os":"linux","arch":"x64"})
    npm WARN tslint@3.15.1 requires a peer of typescript@>=1.7.3 but none was installed.
    npm ERR! Linux 4.7.4-1-ARCH
    npm ERR! argv "/usr/bin/node" "/usr/bin/npm" "install"
    npm ERR! node v6.6.0
    npm ERR! npm  v3.10.8
    npm ERR! code EBADPLATFORM

    npm ERR! notsup Unsupported platform for 7zip-bin-osx@1.0.0: wanted {"os":"darwin","arch":"any"} (current: {"os":"linux","arch":"x64"})
    npm ERR! notsup Valid OS:    darwin
    npm ERR! notsup Valid Arch:  any
    npm ERR! notsup Actual OS:   linux
    npm ERR! notsup Actual Arch: x64

    npm ERR! Please include the following file with any support request:
    npm ERR!     /home/user/some/path/mist-git/npm-debug.log

I'm running `npm install` because I want to run a `gulp mist --platform linux` afterwards. I am not interested in windows or macos binaries. Thanks.

Tail of `npn-debug.log`:

    49079 silly runTopLevelLifecycles Starting
    49080 silly install runPreinstallTopLevelLifecycles
    49081 silly preinstall Mist@0.8.3 /home/user/some/path/mist-git/Mist-b7135eb3
    49082 info lifecycle Mist@0.8.3~preinstall: Mist@0.8.3
    49083 silly lifecycle Mist@0.8.3~preinstall: no script for preinstall, continuing
    49084 silly executeActions Starting
    49085 silly install executeActions
    49086 silly doSerial global-install 0
    49087 silly doParallel fetch 823
    49088 silly rollbackFailedOptional Starting
    49089 silly rollbackFailedOptional Finishing
[...]

PR #1194, opened 2016-09-24
fix and update package.json/npm-shrinkwrap.json

luclu:
fixes https://github.com/ethereum/mist/issues/1188, #1208

npm dependency graph was broken, test this via:
 1. uninstall every falsy globally installed module
 2. `rm -r node_modules`
 3. try to `npm install`

This PR will:
 1. add unmet deps (`tslint@3.15.1`, `typescript@1.7.3`) (npm install --save <module>)
 2. update and cleanup `npm-shrinkwrap.json` via `npm install; npm prune; npm shrinkwrap`
   - changes:
     - update version string to `0.8.3`
     - add both modules (1.) to `package.json`
     - updates shrinkwrap's dependency graph
       - removes `electron-builder` as it is only a development package (`--save-dev`)

Tested on:
 - mac: mac, linux, win builds
 - linux: linux build

PR #1228, opened 2016-09-29
Downloading node client binaries at runtime

hiddentao:
Instead of bundling client binaries at build time we will download then on startup at runtime. This will allow us to update client binaries for all Mist users without having to publish a new Mist release, and it will also reduce the size of Mist/wallet releases. And we will no longer track binaries in the repo (which we should never have done in the first place!). This makes use of [ethereum-client-binaries](https://github.com/hiddentao/ethereum-client-binaries) to facilitate this.

1. It first checks for `ethereumBinaries.json` online in the master repo of Mist project
2. If not found or invalid it will load the one bundled with Mist
3. It will then call into `ethereum-client-binaries` lib with the JSON config and scan for binaries (only Geth supported at the mo)
4. If not matching version of a binary found it will download one and verify again
5. It will then start the app as normal


hiddentao:
Ready for review. Eth is still bundled as normal whilst I work out how best to fetch it on the fly. Gulp script simplified.

luclu:
Works flawlessly - tested on mac and linux :)

Could you also [cleanup your shrinkwrap](https://github.com/ethereum/mist/pull/1194) by running `npm install; npm prune; npm shrinkwrap`?
Without the prune there will be dev-deps included that won't be available under linux (https://github.com/ethereum/mist/issues/1208).
This will also fix the travis build.

hiddentao:
Thanks @luclu , shrinkwrap updated. Also removed some redundant deps.

luclu:
hmm, seems it didn't work as expected, could you repeat it while `rm -r node_modules` before?

hiddentao:
Tried again, after removing `node_modules`

luclu:
Weird, I pushed my `npm install; npm prune; npm shrinkwrap`'s result then..

luclu:
Ok, made travis happy again while on it.

hiddentao:
Thanks @luclu 

With the frequent new releases of Geth that have been happening recently, this PR would really help in rolling out Geth updates for Mist users.

hiddentao:
Annoyingly am getting build errors again. I've done a fresh `npm shrinkwrap` but to no avail. @luclu I think I need your help again.

frozeman:
One problem with your config is that it downloads all files from peters cross builds. We should download the ones on the github release page from there and only take from peter what we don't get there.

@hiddentao if the config file contains the links to the versions to download, how does the client updates the geth nodes automatically, or checks for new versions? Also howe do we restrict updates to only patch versions, as minor amen major versions could contain backwards incompatible changes.

hiddentao:
@frozeman The download links are arbitrary and can be set to whatever. As long as they link to Zips and Tarballs (i.e. no installers) it's fine.

hiddentao:
Parity binaries available here: https://vanity-service.ethcore.io/parity-binaries

luclu:
@hiddentao no problem, I'm at your disposal ;)
> Annoyingly am getting build errors again. I've done a fresh `npm shrinkwrap` but to no avail. @luclu I think I need your help again.

hiddentao:
Thanks @luclu !

evertonfraga:
Testing checklist:

- [ ]  Start mist without geth nodes on folder
	- [ ] With newer geth on remote json
- [ ] Start mist with geth nodes on folder
	- [ ] Newer geth on remote json
- [ ] Start mist with JSON failing to load
- [ ] Test with Sanity check failing on purpose

evertonfraga:
@hiddentao I've raised https://github.com/hiddentao/ethereum-client-binaries/pull/2 in order to make node updates more reliable.

I'll keep testing.

hiddentao:
@evertonfraga Nice one. Only a minor change needed to that PR and it can be merged.

evertonfraga:
Looks really good to me :)
@hiddentao after integrating that PR, we're good to go

hiddentao:
Fixed CI, ready to merge.

PR #1245, opened 2016-10-02
update string as 'v'-prefix was added to version tags

luclu:
As we updated release tags to contain a 'v' as prefix - the current updater-popup will display the version tag as `vv0.8.4`.


Doncharly:
https://kiboplatform.net/DonKibo

https://kiboplatform.net/es/advantages.html

http://huaco2016.wixsite.com/kibo-ingledoncharly

*KIBO. Loto Registracion Only 10 Ethereum Unique Pay*
*Skype Carlos.Silva522*

2016-10-02 7:39 GMT-05:00 Luca Zeug <notifications@github.com>:

PR #1248, opened 2016-10-03
builder: add zip packages

luclu:


evertonfraga:
The idea here is to have both zip and exe for windows?

luclu:
Superseded by https://github.com/ethereum/mist/pull/1276

PR #1266, opened 2016-10-07
remove .editorconfig file

luclu:
@alexvandesande please add/excluded this file locally by using `.git/info/exclude` instead of `.gitignore` (same syntax) **before** merging and pulling this change. (Look here: https://github.com/ethereum/mist/pull/1141/files#r77993385)

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,9 +1,0 @@
-root = true
-
-[*]
-indent_style = space
-indent_size = 4
-end_of_line = lf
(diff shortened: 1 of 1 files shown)


hiddentao:
I actually wanted to check this file into the repo since it makes it easier to ensure everyone is using the same code formatting!

luclu:
I see, http://editorconfig.org. Cool!

PR #1278, opened 2016-10-12
Replace npm shrinkwrap with yarn.lock

hiddentao:
Did a fresh install using yarn and it took <1min. Way faster than npm install.

This replaces shrinkwrap with yarn.lock file so that we'll have more deterministic builds.

Info on yarn -> https://yarnpkg.com/


luclu:
Do we need to adjust travis?

hiddentao:
I suppose we could use yarn within travis to install the deps. Thinking about it, we probably should do that since we no longer use shrinkwrap (which means `npm install` won't be consistent on travis).

hiddentao:
This requires a bit more testing. After deleting `node_modules` and then running `yarn` and then trying `gulp` I got some errors regarding the absence of 7zip-bin-osx. This is an optionally dependency of `7zip-bin`, which is a dep of `electron-builder`. I need to look into it more but it seems that Yarn might not be running through all the install steps NPM does.

luclu:
Yarn Cheatsheet: https://duckduckgo.com/?q=yarn+cheatsheet&ia=cheatsheet&iax=1

luclu:
Yay, starting from today travis natively supports yarn: https://github.com/travis-ci/travis-build/pull/895

luclu:
@hiddentao I fixed the 7zip OS-specific download by adding them as optionalDependencies to the package.json as described in https://yarnpkg.com/en/docs/dependency-types#toc-optionaldependencies.

luclu:
This PR is ready to be merged. 

It includes:
- yarn as the new package manager
  - no more misunderstood `npm-shrinkwrap.json`
  - no problems anymore, just do `yarn` (new commands at [yarn cheatsheet](https://duckduckgo.com/?q=yarn+cheatsheet&ia=cheatsheet&iax=1))
- updated *electron-builder* as there were some problems with deprecated *7zip-modules*
  - updated gulpfile according to minor changes
- travis:
  - native support for yarn
  - enabled yarn cache
  - updated electron-builder dependencies
- tested:
  - [x] mac: all platforms
  - [x] travis: linux
  - [x] linux: linux, win

evertonfraga:
LGTM 👍

PR #1286, opened 2016-10-14
[eslint] added basic ruleset; updated main.js accordingly

luclu:
This PR adds [ESLint](http://eslint.org/) with the [airbnb style guide](https://github.com/airbnb/javascript). 
You can install your editor's eslint plugin to have it check for you while you code. (realtime/on-save)

ESLint brings following advantages among others:
- **consistency of style** across the code base
- **catches syntax errors**/typos
- recommends on **best practices** (e.g. usage of _var/let/const_)
- highlights dead imports/code
- warns of possible vulnerabilities (rather limited; like https://github.com/airbnb/javascript#functions--constructor)

Also the rules will be picked up by code climate on the PR checks.

Following suggested changes were made to `main.js` (sorted by [rule category](http://eslint.org/docs/rules/); this is a subset of the airbnb rules):
- [Fatal] 
  - don't allow return outside of function (was used by squirrel shorthand)
- [Removed]
  - [[keyword-spacing](http://eslint.org/docs/rules/keyword-spacing)] Enforce consistent spacing before and after keywords (e.g. `if (true)` instead of `if(true)`)
- [Possible Errors]
  - [[semi](http://eslint.org/docs/rules/semi)] add missing semicolon (e.g. after return)
- [Strict Mode]
  - [[strict](http://eslint.org/docs/rules/strict)] don't specify `strict mode` (it's [implicitly enabled in node `>=4`](http://node.green/#implicit%20strict%20mode))
- [Best Practices]
  - [[yoda](http://eslint.org/docs/rules/yoda)] disallow yoda (@hiddentao, we could also enforce it :smiling_imp:)
  - [[vars-on-top](http://eslint.org/docs/rules/vars-on-top)] Require Variable Declarations to be at the top of their scope 
  - [[prefer-template](http://eslint.org/docs/rules/prefer-template)] prefer template string over string concatenation
  - [[no-var](http://eslint.org/docs/rules/no-var)] require let or const instead of var (looks like hoisting is dead)
  - [[prefer-const](http://eslint.org/docs/rules/prefer-const)] If a variable is never reassigned, use const 
- [Stylistic Issues]
  - [[comma-dangle](http://eslint.org/docs/rules/comma-dangle)] require trailing comma on last element (note: json does not allow this)
  - [[no-multiple-empty-lines](http://eslint.org/docs/rules/no-multiple-empty-lines)] disallow more than 2 empty following lines 
  - [[newline-after-import](https://github.com/benmosher/eslint-plugin-import/blob/master/docs/rules/newline-after-import.md)] Enforces having an empty line after the last top-level import statement or require call.
  - [[func-names](http://eslint.org/docs/rules/func-names)] Require named function expressions (better for debugging)


hiddentao:
I don't agree with the following "best practices":

* disallow-yoda - _I don't buy the lack of readability argument against this._

* vars-on-top - _in strict mode you have to declare variables before assignment anyway, so why do we need to do this? Plus, if you want to do a `const` you have to do the declaration and assignment in one go._

Would love to hear other people's thoughts on this.

main.js
```diff
@@ -71,112 +64,106 @@ global.webviews = [];
...
+let splashWindow;
+let onReady;
+let startMainWindow;
 
 // This method will be called when Electron has done everything
 // initialization and ready for creating browser windows.
-app.on('ready', function() {
+app.on('ready', () => {
     // if using HTTP RPC then inform user
-    if ('http' === Settings.rpcMode) {
```
evertonfraga: yoda programming

main.js
```diff
@@ -1,36 +1,35 @@
-"use strict";
+const squirrelStartup = require('electron-squirrel-startup');
+// windows only: don't run app during squirrel-install
+function exit() { return; }
```
luclu: This is probably quite controversial.. There is one rule concerning the scope of returns. It basically disallows the use of `return` outside of function scopes/in global the scope. 

This means a construct like:
```
if (true) { return };
```
is not valid under this rule, thus that proxy function `exit()`.

PR #1287, opened 2016-10-15
[refactor] duplicate code in web3admin.js

luclu:
Removes duplicate code.
I'm not sure about the `function(obj) { return obj; }` - quickly replaced it with (the undefined) `obj` or `null` and seems to work.


evertonfraga:
@luclu our code climate score thanks you for that.

luclu:
@evertonfraga this was a thankful PR, picking that low-hanging fruit ;)

PR #1290, opened 2016-10-15
update geth to 1.4.18

luclu:
uploaded!


luclu:
The links resolve and download proper using wget.

PR #1326, opened 2016-10-19
[travis] add gitter notification webhook

luclu:
- compare go-ethereum: [.travis.yml#L40-L45](https://github.com/ethereum/go-ethereum/blob/develop/.travis.yml#L40-L45)
- env var already set in travis

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -27,3 +27,9 @@ script:
 notifications:
   email:
     - EMAIL
+  webhooks:
+    urls:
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -27,3 +27,9 @@ script:
 notifications:
   email:
     - EMAIL
+  webhooks:
+    urls:
+      - YOUR_WEBHOOK_URL
```
luclu: I already set the envs in travis.

PR #1348, opened 2016-10-25
Add SHA-256 hashes for current node binary packages

hiddentao:


hiddentao:
#1349 redoes this against develop

PR #1360, opened 2016-10-28
Adding security overlay on Mist window when opening sensitive popups

evertonfraga:
This PR adds an overlay and blurs Mist main window when `sendTransactionConfirmation` appears.

WIP.


evertonfraga:
Expected behavior:
image

There's also a [video demo](https://www.dropbox.com/s/2fokfn3yu85zr1w/overlay.mov?dl=0) on Mac.

alexvandesande:
Our current modals are light white, not dark, so in terms of consistency, we should have these be the same

screen shot 2016-10-31 at 15 34 18

evertonfraga:
Ok @alexvandesande, I'll perform the changes. worth noticing I'm not blurring on linux, due to performance issues and historical user problem with graphic cards.

alexvandesande:
@evertonfraga sure. Then on linux it would be just a white overlay ;-) 

Of course we could try a dark blur on the wallet, but my past experience tells it doesn't look good

PR #1382, opened 2016-11-04
Add 0 to addresses in parameters

alexvandesande:


frozeman:
screen shot 2016-11-04 at 17 45 05

modules/abi.js
```diff
@@ -11,27 +11,36 @@ function isHexType(type) {
...
-            var conversionFlag = isHexType(type) ? 'hex' : null,
+            let conversionFlag = isHexType(type) ? 'hex' : null,
                 prefix = isHexType(type) ? '0x' : '';
 
             paramsResponse[index] = paramsResponse[index].toString(conversionFlag);
 
+            if(type === 'address')
+            	paramsResponse[index] = padLeft(paramsResponse[index], 40);
+            else if(type.match(/bytes\d+/g))
+            	paramsResponse[index] = padLeft(paramsResponse[index], 64);
```
alexvandesande: shoudn't this depend on the size of the bytes? 64 chars is only for bytes256

frozeman: normally parameters are always in sizes of 32bytes chunks, we could also leave that out, but it makes it more consistent
Though we should add a "title" to the dapp_output in the elements package

alexvandesande:
Now it correctly detects the correct length:
screen shot 2016-11-04 at 18 06 37

Also, if it goes over the line, :hover enables line break:
screen shot 2016-11-04 at 18 07 07

PR #1383, opened 2016-11-04
Client binary refactor - asking user for permission before updating Geth

hiddentao:
How it works:

* When app starts it checks for new remote config JSON
* If also checks for locally saved config JSON
* If both match then app continues as normal
* If there is a mismatch then:
  * If local config is empty then this is a first-time run so app simply saves the remote to local and then downloads the Geth client straight away
  * If local config is non-empty then this means a new Geth update is out. The app will show a popup window to user asking if they wish to upgrade. If user accept then the app will save the remote to local (overwriting the existing local) and restart itself


hiddentao:
This is ready for review.

modules/clientBinaryManager.js
```diff
@@ -3,29 +3,43 @@
...
 const Settings = require('./settings');
+const Windows = require("./windows");
 const ClientBinaryManager = require('ethereum-client-binaries').Manager;
 const EventEmitter = require('events').EventEmitter;
 
 const log = require('./utils/logger').create('ClientBinaryManager');
 
 
+const ALLOWED_DOWNLOAD_URLS_REGEX =
+    /^http[s]?:\/\/([^\.]*\.)?ethereum\.org\/(.*)?|^http[s]?:\/\/([^\.]*\.)?bintray\.com\/karalabe\/ethereum\/(.*)?/;
```
alexvandesande: This pattern doesn't match the current urls as the correct bintray url is bintray.com/*artifact/download*/karalabe.. This one should do the trick:

`
^http[s]?:\/\/([^\.]*\.)?ethereum\.org\/(.*)?|^http[s]?:\/\/([^\.]*\.)?bintray\.com\/artifact\/download\/karalabe\/ethereum\/(.*)?
`

Otherwise it's correct.

alexvandesande:
superseded by https://github.com/ethereum/mist/pull/1392

Issue #1402, opened 2016-11-10
Updating meteor breaks mist 0.8.7


giacecco:
#### System information
``````
Version: `0.8.7 from sources`
OS & Version: `linux (Fedora 24)`
Node type: `geth`
``````

Starting meteor in mist's ```interface``` folder suggests upgrading it...

```
[giacecco@giaceccos-linux interface]$ meteor
[[[[[ ~/Downloads/mist-0.8.7/interface ]]]]]  

=> Started proxy.                             
=> Started MongoDB.                           
=> Meteor 1.4.2.1 is available. Update this project with 'meteor update'.
=> Started your app.              
(...)
```

... but if you do, it causes https://github.com/ethereum/mist/issues/1401 . 

Reinstalling from scratch fixes it.

The full meteor upgrade process is shown at https://gist.github.com/giacecco/44547b06c585592f21846970ba4667fe in case it can suggest what exact component does the damage.

frozeman:
Thanks for the gist.
You shouldn't update meteor yourself, it will surely break some dependencies

PR #1406, opened 2016-11-10
Added note not to upgrade Meteor to avoid breaking Mist

giacecco:
To mitigate https://github.com/ethereum/mist/issues/1402 .


luclu:
Thanks @giacecco 

Apparently there is also an ENV that could be leveraged to suppress those notifications.
Ironically it is only available in meteor `>=1.4.1`. https://github.com/meteor/meteor/issues/7026#issuecomment-242025235

pinging @frozeman is there any path to upgrade outlined yet?

frozeman:
We updated meteor. Thanks anyway

PR #1414, opened 2016-11-14
ETH-01-cure fixes

frozeman:
Fixes currently:
- ETH-01-002
- ETH-01-003
- ETH-01-005

Questions here:
The question is do we want to avoid `data://` URIs and what threat do they bring.

Note that these URIs are only prevent in the browserbar and as a possible full page load. Not as assets in a webpage


interface/client/lib/helpers/helperFunctions.js
```diff
@@ -50,6 +50,21 @@ Helpers.formatUrl = function(url){
...
+@method sanatizeUrl
+@param {String} url
+**/
+Helpers.sanatizeUrl = function(url){    
+    // add http:// if no protocol is present
+    if(_.isString(url)) {
+        url = url.replace(/^(javascript:)?(data:)?\/\//i, 'http:');
+    }
+
+    return url;
```
holiman: blacklist is not sufficient. `javaScript://` is also valid, and maybe special characters such as %0A . 

Also, why only handle strings? 
If you expect anything other than strings, such as location-objects, you should filter them. If you expect string input, then cast the url to string and do the sanitation unconditionally.

holiman: Sorry, I missed the `/i` . But javascript can still be broken up with tab or %0A. I'd rather see: 

1. Robust URL parsing, either using DOM anchor element to parse URL or a proper URL parsing library. 
2. Whitelist, only allowing the protocols we want to allow, instead of blacklisting a few.

frozeman: I will look into a proper JS library. Currently the URL is never a location object, but i can add checks for them too.

Whitelisting would make sense if we fully know what people will use and currently we don't want to restrict unknown URIs. E.g. ftp, etc.
As it looks only JavaScript is semi-dangerous as it can spawn ui actions, but this js has also not added privileges than the webpage itself.
So removing JavaScript seems to be the most dangerous one, can you think of any other?

frozeman: @holiman i changed it now to use the parser object form your gist, but concerning the whitelisting im not yet convinced if we are not blocking uses cases. This current sanatizer should prevent spoofing. We can discuss in the future if we want to add more

evertonfraga:
Testing
- [ ] ETH-01-002 
- [ ] ETH-01-003 
- [ ] ETH-01-005

interface/client/lib/helpers/helperFunctions.js
```diff
@@ -50,6 +50,28 @@ Helpers.formatUrl = function(url){
...
+    var parser;
+
+    if(_.isString(url)) {
+        parser = document.createElement('a');
+        parser.href = url;
+    } else if(_.isObject(url) && url.protocol) {
+        parser = url;
+    }
+
+    parser.protocol = parser.protocol.replace(/^(javascript:)?(data:)?(:)?/i, 'http:');
```
holiman: See [playground example](aW5pdEFwcCh3aW5kb3cpO2FkZENvZGVNaXJyb3IodW5lc2NhcGUoJ3NhbmF0aXplVXJsJTIwJTNEJTIwZnVuY3Rpb24lMjh1cmwlMjklN0IlMEElMjAlMjAlMjAlMjAlMjB2YXIlMjBwYXJzZXIlM0IlMEElMjAlMjAlMjAlMjAlMjBzaG93JTI4JTIyVVJMJTNBJTIyK3VybCUyOSUwQSUyMCUyMCUyMCUyMCUyMCUwQSUyMCUyMCUyMCUyMCUyMGlmJTI4dHlwZW9mJTI4dXJsJTI5JTNEJTNEJTI3c3RyaW5nJTI3JTI5JTIwJTdCJTBBJTIwJTIwJTIwJTIwJTIwJTIwJTIwJTIwJTIwcGFyc2VyJTIwJTNEJTIwZG9jdW1lbnQuY3JlYXRlRWxlbWVudCUyOCUyN2ElMjclMjklM0IlMEElMjAlMjAlMjAlMjAlMjAlMjAlMjAlMjAlMjBwYXJzZXIuaHJlZiUyMCUzRCUyMHVybCUzQiUwQSUyMCUyMCUyMCUyMCUyMCU3RCUyMGVsc2UlMjBpZiUyOF8uaXNPYmplY3QlMjh1cmwlMjklMjAlMjYlMjYlMjB1cmwucHJvdG9jb2wlMjklMjAlN0IlMEElMjAlMjAlMjAlMjAlMjAlMjAlMjAlMjAlMjBwYXJzZXIlMjAlM0QlMjB1cmwlM0IlMEElMjAlMjAlMjAlMjAlMjAlN0QlMEElMjAlMDklMjAlMEElMjAlMjAlMDlzaG93JTI4JTIyJTVCMiU1RCUyMEJlZm9yZSUyMHJlcGxhY2UlM0ElMjAlMjIrcGFyc2VyJTI5JTBBJTIwJTIwJTIwJTIwJTIwcGFyc2VyLnByb3RvY29sJTIwJTNEJTIwcGFyc2VyLnByb3RvY29sLnJlcGxhY2UlMjgvJTVFJTI4amF2YXNjcmlwdCUzQSUyOSUzRiUyOGRhdGElM0ElMjklM0YlMjglM0ElMjklM0YvaSUyQyUyMCUyN2h0dHAlM0ElMjclMjklM0IlMEElMjAlMjAlMDlzaG93JTI4JTIyJTVCMyU1RCUyMEFmdGVyJTIwcmVwbGFjZSUzQSUyMiUyMCslMjBwYXJzZXIlMjklMEElMjAlMjAlMjAlMjAlMjAlMEElMjAlMjAlMjAlMjAlMjB1cmwlMjAlM0QlMjBwYXJzZXIudG9TdHJpbmclMjglMjklM0IlMEElMjAlMEElMjAlMjAlMjAlMjAlMjByZXR1cm4lMjB1cmwlM0IlMEElMjAlN0QlM0IlMEElMEFzaG93JTI4c2FuYXRpemVVcmwlMjglMjJodHRwcyUzQS8vd3d3LmRuLnNlJTIyJTI5JTI5JykpOw0KdXBkYXRlT3V0cHV0KCk7DQo). This function will replace "https://" urls with "http://"

This is (I think) because the matches all, due to there only being optional capture groups. See [here](aW5pdEFwcCh3aW5kb3cpO2FkZENvZGVNaXJyb3IodW5lc2NhcGUoJ3NhbmF0aXplVXJsJTIwJTNEJTIwZnVuY3Rpb24lMjh1cmwlMjklN0IlMEElMjAlMjAlMjAlMjAlMjB2YXIlMjBwYXJzZXIlM0IlMEElMjAlMjAlMjAlMjAlMjBzaG93JTI4JTIyVVJMJTNBJTIyK3VybCUyOSUwQSUyMCUyMCUyMCUyMCUyMCUwQSUyMCUyMCUyMCUyMCUyMGlmJTI4dHlwZW9mJTI4dXJsJTI5JTNEJTNEJTI3c3RyaW5nJTI3JTI5JTIwJTdCJTBBJTIwJTIwJTIwJTIwJTIwJTIwJTIwJTIwJTIwcGFyc2VyJTIwJTNEJTIwZG9jdW1lbnQuY3JlYXRlRWxlbWVudCUyOCUyN2ElMjclMjklM0IlMEElMjAlMjAlMjAlMjAlMjAlMjAlMjAlMjAlMjBwYXJzZXIuaHJlZiUyMCUzRCUyMHVybCUzQiUwQSUyMCUyMCUyMCUyMCUyMCU3RCUyMGVsc2UlMjBpZiUyOF8uaXNPYmplY3QlMjh1cmwlMjklMjAlMjYlMjYlMjB1cmwucHJvdG9jb2wlMjklMjAlN0IlMEElMjAlMjAlMjAlMjAlMjAlMjAlMjAlMjAlMjBwYXJzZXIlMjAlM0QlMjB1cmwlM0IlMEElMjAlMjAlMjAlMjAlMjAlN0QlMEElMjAlMDklMjAlMEElMjAlMjAlMDlzaG93JTI4JTIyJTVCMiU1RCUyMEJlZm9yZSUyMHJlcGxhY2UlM0ElMjAlMjIrcGFyc2VyJTI5JTBBJTIwJTIwJTIwJTIwJTBBJTIwJTIwJTIwJTIwaWYlMjglMjBwYXJzZXIucHJvdG9jb2wlMjAlM0QlM0QlMjAlMjJqYXZhc2NyaXB0JTNBJTIyJTIwJTdDJTdDJTBBJTIwJTIwJTIwJTIwJTIwJTIwJTIwcGFyc2VyLnByb3RvY29sJTIwJTNEJTNEJTIyZGF0YSUzQSUyMiUyOSUwQSUyMCUyMCUyMCUyMCU3QiUwQSUyMCUyMCUyMCUyMCUyMCUyMHBhcnNlci5wcm90b2NvbCUyMCUzRCUyMCUyMmh0dHBzJTNBLy8lMjIlMEElMjAlMjAlMjAlMjAlN0QlMEElMEElMjAlMjBzaG93JTI4JTIyJTVCMyU1RCUyMEFmdGVyJTIwcmVwbGFjZSUzQSUyMiUyMCslMjBwYXJzZXIlMjklMEElMjAlMjAlMjAlMjAlMjAlMEElMjAlMjAlMjAlMjAlMjB1cmwlMjAlM0QlMjBwYXJzZXIudG9TdHJpbmclMjglMjklM0IlMEElMjAlMEElMjAlMjAlMjAlMjAlMjByZXR1cm4lMjB1cmwlM0IlMEElMjAlN0QlM0IlMEElMEFzaG93JTI4c2FuYXRpemVVcmwlMjglMjJqYXZhcyU1Q3RjcmlwVCUzQWFsZXJ0JTI4MSUyOSUyMiUyOSUyOScpKTsNCnVwZGF0ZU91dHB1dCgpOw0K) for a more robust example.


And I still think that any input _not_ explicitly  supported, should be coerced into a string. Otherwise it does not get filtered.

holiman: Here's an [example](http://martin.swende.se/tools/inject_it.js/#aW5pdEFwcCh3aW5kb3cpO2FkZENvZGVNaXJyb3IodW5lc2NhcGUoJ3NhbmF0aXplVXJsJTIwJTNEJTIwZnVuY3Rpb24lMjh1cmwlMjklN0IlMEElMjAlMjAlMjAlMjB2YXIlMjBwYXJzZXIlMjAlM0QlMjBkb2N1bWVudC5jcmVhdGVFbGVtZW50JTI4JTI3YSUyNyUyOSUzQiUwQSUyMCUyMCUyMCUyMHBhcnNlci5ocmVmJTIwJTNEJTIwJTIyJTIyK3VybCUzQiUyMC8vU3RyaW5naWZ5JTIwVVJMJTBBJTIwJTA5JTBBJTIwJTIwJTIwJTIwaWYlMjglMjBwYXJzZXIucHJvdG9jb2wlMjAlM0QlM0QlMjAlMjJqYXZhc2NyaXB0JTNBJTIyJTIwJTdDJTdDJTBBJTIwJTIwJTIwJTIwJTIwJTIwJTIwcGFyc2VyLnByb3RvY29sJTIwJTNEJTNEJTIwJTIyZGF0YSUzQSUyMiUyOSUwQSUyMCUyMCUyMCUyMCU3QiUwQSUyMCUyMCUyMCUyMCUyMCUyMHBhcnNlci5wcm90b2NvbCUyMCUzRCUyMCUyMmh0dHBzJTNBLy8lMjIlMEElMjAlMjAlMjAlMjAlN0QlMEElMjAlMjAlMjAlMjAlMjByZXR1cm4lMjBwYXJzZXIudG9TdHJpbmclMjglMjklM0IlMEElMjAlN0QlM0IlMEElMEFzaG93JTI4c2FuYXRpemVVcmwlMjglMjJqYXZhcyU1Q3RjcmlwVCUzQWFsZXJ0JTI4MSUyOSUyMiUyOSUyOScpKTsNCnVwZGF0ZU91dHB1dCgpOw0K) which is even more robust, although I'd still prefer whitelisting protos instead of blacklisting

frozeman: I tested you version and they didn't cover everything.
I found the best way which is here: Here's a link to share:
http://martin.swende.se/tools/inject_it.js/#aW5pdEFwcCh3aW5kb3cpO2FkZENvZGVNaXJyb3IodW5lc2NhcGUoJyUwQSUwQS8vJTIwYmVzdCUwQXNhbmF0aXplVXJsJTIwJTNEJTIwZnVuY3Rpb24lMjh1cmwlMjklN0IlMEElMjAlMjAlMjAlMjB1cmwlMjAlM0QlMjBTdHJpbmclMjh1cmwlMjklM0IlMEElMEElMjAlMjAlMjAlMjB1cmwlMjAlM0QlMjB1cmwucmVwbGFjZSUyOC8lNUIlNUN0JTVDbiU1Q3IlNUNzJTVEKy9nJTJDJTIwJTI3JTI3JTI5JTNCJTBBJTIwJTIwdXJsJTIwJTNEJTIwdXJsLnJlcGxhY2UlMjgvJTVFJTI4JTI4JTNGJTNBamF2YXNjcmlwdCUyOSUzRiUyOCUzRiUzQWRhdGElMjklM0YlNUIlM0ElNUMvJTVDLyU1RCU3QjElMkMzJTdEJTI5L2klMkMlMjAlMjdodHRwJTNBLy8lMjclMjklM0IlMEElMEElMjAlMjAlMjAlMjByZXR1cm4lMjB1cmwlM0IlMEElMjAlN0QlM0IlMEElMEEvKiUwQXNhbmF0aXplVXJsJTIwJTNEJTIwZnVuY3Rpb24lMjh1cmwlMjklN0IlMEElMjAlMjAlMjAlMjB2YXIlMjBwYXJzZXIlMjAlM0QlMjBkb2N1bWVudC5jcmVhdGVFbGVtZW50JTI4JTI3YSUyNyUyOSUzQiUwQSUyMCUyMCUyMCUyMHBhcnNlci5ocmVmJTIwJTNEJTIwJTIyJTIyK3VybCUzQiUyMC8vU3RyaW5naWZ5JTIwVVJMJTBBJTIwJTA5JTBBJTIwJTIwJTIwJTIwaWYlMjglMjBwYXJzZXIucHJvdG9jb2wlMjAlM0QlM0QlMjAlMjJqYXZhc2NyaXB0JTNBJTIyJTIwJTdDJTdDJTBBJTIwJTIwJTIwJTIwJTIwJTIwJTIwcGFyc2VyLnByb3RvY29sJTIwJTNEJTNEJTIwJTIyZGF0YSUzQSUyMiUyOSUwQSUyMCUyMCUyMCUyMCU3QiUwQSUyMCUyMCUyMCUyMCUyMCUyMHBhcnNlci5wcm90b2NvbCUyMCUzRCUyMCUyMmh0dHBzJTNBLy8lMjIlMEElMjAlMjAlMjAlMjAlN0QlMEElMjAlMjAlMjAlMjAlMjByZXR1cm4lMjBwYXJzZXIudG9TdHJpbmclMjglMjklM0IlMEElMjAlN0QlM0IlMEElMjAlMEElMjBzYW5hdGl6ZVVybCUyMCUzRCUyMGZ1bmN0aW9uJTI4dXJsJTI5JTdCJTBBJTIwJTIwJTIwJTIwdmFyJTIwcGFyc2VyJTIwJTNEJTIwZG9jdW1lbnQuY3JlYXRlRWxlbWVudCUyOCUyN2ElMjclMjklM0IlMEElMjAlMjAlMjAlMjBwYXJzZXIuaHJlZiUyMCUzRCUyMCUyMiUyMit1cmwlM0IlMjAvL1N0cmluZ2lmeSUyMFVSTCUwQSUyMCUwOSUwQSUyMCUyMCUyMCUyMGlmJTI4JTIwcGFyc2VyLnByb3RvY29sJTIwJTNEJTNEJTIwJTIyamF2YXNjcmlwdCUzQSUyMiUyMCU3QyU3QyUwQSUyMCUyMCUyMCUyMCUyMCUyMCUyMHBhcnNlci5wcm90b2NvbCUyMCUzRCUzRCUyMCUyMmRhdGElM0ElMjIlMjklMEElMjAlMjAlMjAlMjAlN0IlMEElMjAlMjAlMjAlMjAlMjAlMjBwYXJzZXIucHJvdG9jb2wlMjAlM0QlMjAlMjJodHRwcyUzQS8vJTIyJTBBJTIwJTIwJTIwJTIwJTdEJTBBJTIwJTIwJTIwJTIwJTIwcmV0dXJuJTIwcGFyc2VyLnRvU3RyaW5nJTI4JTI5JTNCJTBBJTIwJTdEJTNCJTBBJTIwJTBBJTIwKi8lMEElMEFzaG93JTI4c2FuYXRpemVVcmwlMjglMjJqYXZhcyU1Q3RjcmlwVCUzQWFsZXJ0JTI4MSUyOSUyMiUyOSUyOSUwQXNob3clMjglMjItLS0tJTIyJTI5JTBBc2hvdyUyOHNhbmF0aXplVXJsJTI4JTIyaHR0cCUzQS8vYWxlcnQlMjgxJTI5JTIyJTI5JTI5JTBBc2hvdyUyOCUyMi0tLS0lMjIlMjklMEFzaG93JTI4c2FuYXRpemVVcmwlMjglMjJodHRwcyUzQS8vdGVzdC5jb20lMjIlMjklMjklMEFzaG93JTI4JTIyLS0tLSUyMiUyOSUwQXNob3clMjhzYW5hdGl6ZVVybCUyOCUyMmphdmFzJTVDdGNyaXBUJTNBLy8lMjUwQWFsZXJ0JTI4MSUyOSUyMiUyOSUyOSUwQXNob3clMjglMjItLS0tJTIyJTI5JTBBc2hvdyUyOHNhbmF0aXplVXJsJTI4JTIyZmlsZSUzQS8vdGVzdC5jb20lMjIlMjklMjklMEFzaG93JTI4JTIyLS0tLSUyMiUyOSUwQXNob3clMjhzYW5hdGl6ZVVybCUyOCUyMmZ0cCUzQS8vdGVzdC5jb20lMjIlMjklMjklMEFzaG93JTI4JTIyLS0tLSUyMiUyOSUwQXNob3clMjhzYW5hdGl6ZVVybCUyOCUyMi8vdGVzdC5jb20lMjIlMjklMjknKSk7DQp1cGRhdGVPdXRwdXQoKTsNCg==

```js
sanatizeUrl = function(url){
// make sure location objects are string too
    url = String(url);

// remove whitespaces and tabs
    url = url.replace(/[\t\n\r\s]+/g, '');
// replace javascript:// data:// and  //
  url = url.replace(/^((?:javascript)?(?:data)?[:\/\/]{1,3})/i, 'http://');

    return url;
 };
```

holiman: What specifically did not work with the DOM-anchor method ? 
I like that you now force it to be a string, but dislike that it's gone back to regexp. URL parsing is by no means trivial, so I'd still advocate a proper URL parsing library if the DOM is insufficient.

interface/client/lib/helpers/helperFunctions.js
```diff
@@ -50,6 +50,21 @@ Helpers.formatUrl = function(url){
...
+Sanatizes URLs to prevent phishing and XSS attacks
+
+@method sanatizeUrl
+@param {String} url
+**/
+Helpers.sanatizeUrl = function(url){
+    url = String(url);
+
+    url = url.replace(/[\t\n\r\s]+/g, '');
+    url = url.replace(/^((?:javascript)?(?:data)?[:\/\/]{1,3})/i, 'http://');
```
alexvandesande: Shouldn't we do the opposite? Have only some allowed prefixes and remove all the rest?

interface/client/lib/helpers/helperFunctions.js
```diff
@@ -50,6 +50,21 @@ Helpers.formatUrl = function(url){
 };
 
 /**
+Sanatizes URLs to prevent phishing and XSS attacks
+
+@method sanatizeUrl
+@param {String} url
+**/
+Helpers.sanatizeUrl = function(url){
```
alexvandesande: It's "san*i*tize"

interface/client/templates/views/tab.js
```diff
@@ -50,14 +49,24 @@ Template['views_tab'].onRendered(function(){
         // Meteor.clearTimeout(timeoutId);
         TemplateVar.set(template, 'loading', false);
 
+        var titleFull = webview.getTitle(),
+            title;
+
+        if(titleFull && titleFull.length > 40) {
```
alexvandesande: Why are you cropping it on javascript? This could be accomplished with css, using `overflow:hidden; text-overflow: ellipsis; white-space:nowrap;`. Word width is not the same as character count.

modules/preloader/mistUI.js
```diff
@@ -5,7 +5,7 @@
 require('./include/common')('mist');
 const { ipcRenderer: ipc, remote, webFrame } = require('electron');
 const { Menu, MenuItem } = remote;
-const syncDb = require('../syncDb.js');
+const dbSync = require('../dbSync.js');
```
alexvandesande: what's the difference between syncDB and dbSync?

modules/ipc/ipcProviderBackend.js
```diff
@@ -337,14 +337,14 @@ class IpcProviderBackend {
         })
         .catch((err) => {
 
-            err = this._makeErrorResponsePayload(payload || {}, {
+            var returnErr = this._makeErrorResponsePayload(payload || {}, {
```
alexvandesande: use `let` instead?

modules/preloader/mistUI.js
frozeman: Its just a name refactor. The reason is that now the files db.js and dbSync.js are next to each other, which helps with mental grouping.
Also we calle the nodeSync: nodeSync ;) so consistency.

This has actually northie todo with the vu.

modules/ipc/ipcProviderBackend.js
frozeman: ok

frozeman: But thats actually part of another PR: https://github.com/ethereum/mist/pull/1421 i will change it there

interface/client/templates/views/tab.js
frozeman: The point is we don't want to add the tex overflow on any element where we use that. 
Titles more than 40 characters are uncommon and for that i added the property "nameFull"

interface/client/lib/helpers/helperFunctions.js
frozeman: There was a discussion above already. My point is that there are a lot of URI types and we don't want to restrict use cases. "Javascript:" is the most dangerous one. But even this cannot do any harm, besides annoy people or mislead them with alert windows.

interface/client/lib/helpers/helperFunctions.js
```diff
@@ -50,6 +50,21 @@ Helpers.formatUrl = function(url){
...
+Sanatizes URLs to prevent phishing and XSS attacks
+
+@method sanitizeUrl
+@param {String} url
+**/
+Helpers.sanitizeUrl = function(url){
+    url = String(url);
+
+    url = url.replace(/[\t\n\r\s]+/g, '');
+    url = url.replace(/^((?:javascript)?(?:data)?[:\/\/]{1,3})/i, 'http://');
```
holiman: As I've said, I dislike using naive regexps for parsing URLs. I don't have a bypass, but here's a simple false positive:

sanitizeUrl("javascript/jquery.js"); -> http://jquery.js

So if it's ever used to filter/sanitize relative URLs, it will fail if for `data` and `javascript` folders.

PR #1420, opened 2016-11-15
[ESLint] modules: simple style-related issues

luclu:
This is the first of two PRs fixing the remaining ESLint issues in the `modules` directory.
It only contains changes that are not likely to break anything.
The second PR will contain smaller changes where some of those will require testing.


modules/abi.js
```diff
@@ -11,42 +12,41 @@ function isHexType(type) {
...
-            	paramsResponse[index] = padLeft(paramsResponse[index], Number(res[1])*2);
+            const res = type.match(/bytes(\d+)/i);
+            if (type === 'address') {
+                paramsResponse[index] = padLeft(paramsResponse[index], 40);
+            } else if (res) {
+                paramsResponse[index] = padLeft(paramsResponse[index], Number(res[1]) * 2);
+            }
 
-            paramsDictArr.push({type: type, value: prefix + paramsResponse[index]});
+            paramsDictArr.push({ type, value: prefix + paramsResponse[index] });
```
evertonfraga: ⚠️ marking for proper testing

luclu: This is valid in ES 6 -> [Literal Property Value Shorthand](https://ariya.io/2013/02/es6-and-object-literal-property-value-shorthand). One may specify only the literal if key and value are similar.

PR #1422, opened 2016-11-15
Eth 01 cure

frozeman:
Fixes:

- ETH-01-007
- ETH-01-008
- ETH-01-009
- ETH-01-012
- ETH-01-014

~Preventing link click-redirects from within tabs is currently not possible due to a pending issue in electron: https://github.com/electron/electron/issues/8008~

One issue: If a web content is selected (clicked in a webview), then file dropping doesn't work anymore.
TODO: We can think about re-activating it or blocking it completely.

- ETH-01-014 needs to be retested on a windows system! @evertonfraga @luclu ?

for ETH 008 and 002 see:
screen shot 2016-11-15 at 19 33 15


Manupri:
Hello 
i Have request to the Wallet 0xB7DE2840470bCC5591F6bf0B8e048ad16BC8bcc2

13.5 ETH

I not Have the Ether in My Wallet

Thanks

frozeman:
@Manupri this is a pull request, this is not for issues. Please create an issue here: https://github.com/ethereum/mist/issues

Concerning your ether, they are at this address: https://etherscan.io/address/0xB7DE2840470bCC5591F6bf0B8e048ad16BC8bcc2
Maybe you have to sync you chain fully.

alexvandesande:
This does not solve https://github.com/ethereum/security-issues/issues/22 yet. I am testing and still can see it.

frozeman:
This PR is also solving the flickering of icons issue, which was cause by using ``{{> elements_img }}`` which fades images in when loaded. As the tabs where changed often and meteor seemed to "reload" the templates all the time, this lead to the flickering.

frozeman:
@alexvandesande  The last PR also removed the behaviour where when you clicked in the browser bar, it selected the whole text.
This made it impossible to select parts of the URL.
If we want to have the one click connivence we have to come up with a way which doesn't block selecting text.

interface/client/lib/helpers/helperFunctions.js
```diff
@@ -50,6 +50,21 @@ Helpers.formatUrl = function(url){
...
 /**
+Sanatizes URLs to prevent phishing and XSS attacks
+
+@method sanatizeUrl
+@param {String} url
+**/
+Helpers.sanatizeUrl = function(url){    
+    // add http:// if no protocol is present
+    if(_.isString(url)) {
+        url = url.replace(/^(javascript\:)?(data\:)?\/\//i,'http:');
```
luclu: The regex doesn't seem to be valid. It defines two capture groups but won't make use of them. Thus the `replace` will omit the `//` and output: `http:victim.com/superlegit/...`.

It currently looks like this:

bildschirmfoto 2016-11-17 um 23 58 41

Instead we might use this regex:

bildschirmfoto 2016-11-17 um 23 58 54

`url = url.replace(/^(?:javascript|data)(?=:\/\/)/i, 'http');`

Note:
 - `(?:group)` will make it non-capture as it is not needed
 - `foo(?=bar)` is a positive lookahead and will only pass "foo" to the replace function

frozeman:
Can everybody please test now all issues again? :)
I deactivated the mist:// protocol for now, as it was never actually meant to be used yet

frozeman:
014 needs to be retested on a windows system!

interface/client/collections.js
```diff
@@ -8,26 +8,17 @@
...
 }
 
 
-// Contains the address book
-AddressBook = new Mongo.Collection('address-book', {connection: null});
-new PersistentMinimongo2(AddressBook, 'Mist');
-
-
-// Contains the accounts
-DoogleLastVisitedPages = new Mongo.Collection('doogle-last-visted-pages', {connection: null});
```
alexvandesande: Finally getting rid of this horrible variable name! 😜

PR #1436, opened 2016-11-20
update geth 1.5.2 download links

luclu:
Change [update URL](https://github.com/ethereum/mist/blob/develop/modules/clientBinaryManager.js#L54) to `https://raw.githubusercontent.com/ethereum/mist/luclu_update-geth-1.5.2/clientBinaries.json` to test. 

Tested:
- [x] linux64
- [x] mac
- [x] win64
- [x] win32


frozeman:
@evertonfraga can you test the rest?

frozeman:
If they only provide MD5, lets add a md5 check.

luclu:
*Copied from the chat:*

Mah win isn’t working yet. While `unpacking geth-windows-amd64-1.5.2-c8695209\\geth.exe` it should extract `geth.exe` but instead creates a new folder called `dows-amd64-1.5.2-c8695209`.
The first missing characters `geth-win` have the same length as `geth.exe.` ;P

Hmm, this behaviour is caused by missing error handling at the unzip statement. unzip fails with this error on both windows zips:
```
Error: invalid signature: 0x80014
```
As a migration path we can upload new non-problematic zips to bintray/github.
Here is a basic script to trigger the error:
```
var fs = require('fs');
var unzip = require('unzip');

fs.createReadStream('geth-windows-386-1.5.2-c8695209.zip').pipe(unzip.Extract({ path: 'output' }));
```
Apperently this is a 2 years old issue. There also is a fork fixing it as the packing won’t receive update anymore: https://github.com/glebdmitriew/node-unzip-2 - this works.

frozeman:
Why closing this?

evertonfraga:
@frozeman this is automatic behavior from github. Ram merged the issue on ethereum-client-binaries, then GH thinks it has to be closed here.

PR #1444, opened 2016-11-21
[editorconfig] change json indent from 2 to 4 spaces

luclu:


frozeman:
But you are changing it here from 4 to 2, as far as i see.

evertonfraga:
@frozeman json's default is 4 spaces. luclu removed the override that changed it to 2 only.

frozeman:
Lets discuss this in gitter.

PR #1445, opened 2016-11-21
[cleanup] remove obsolete codesigning files

luclu:


frozeman:
Why are they obsolete??
These are a work in progress and probably still necessary for a future proper signing. Until we don't have a proper signing, i would keep them, as they were my WIP on that front.

luclu:
The signing process in electron-builder differs from xcode. Those files are not needed as any necessary files will be auto-generated and restructured into the final `.app`.

@frozeman Code-Signing works with my developer certificate.

Also please note that the production certificate can only be used on the MacStore and will be auto-signed by Apple in the process. Please see [MaintainingCertificates](https://developer.apple.com/library/content/documentation/IDEs/Conceptual/AppDistributionGuide/MaintainingCertificates/MaintainingCertificates.html#//apple_ref/doc/uid/TP40012582-CH31-SW41).

frozeman:
If the signing using electron builder works we can rid of these for now. We still have them in the hortatory if needed

PR #1447, opened 2016-11-21
update geth to 1.5.2

luclu:


evertonfraga:
it looks like it has finished a while ago, but didn't signaled back to github.

https://travis-ci.org/ethereum/mist/builds/177758582

PR #1454, opened 2016-11-23
[UX] Adds shortcuts to select tabs: Cmd + [1-9]

evertonfraga:
Simple like that. The absence of those shortcuts was sort of bugging me.

### DIFF
--- a/interface/client/lib/helpers/helperFunctions.js
+++ b/interface/client/lib/helpers/helperFunctions.js
@@ -68,7 +68,7 @@ Format Urls, e.g add a default protocol if on is missing.
 @method formatUrl
 @param {String} url
 **/
-Helpers.formatUrl = function(url){    
+Helpers.formatUrl = function(url){
     // add http:// if no protocol is present
     if(url && url.indexOf('://') === -1)
         url = 'http://'+ url;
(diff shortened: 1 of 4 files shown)


frozeman:
I think this code is better part of the sideBar.js Can you please move it over?

evertonfraga:
@frozeman yes, great suggestion. I've moved to sidebar and extracted the two methods to Helpers.

alexvandesande:
It's not working for me. Actually, it's not even invoking the 'document keydown' event, I tested and it won't even do a console.log

Also, why is command+9 a shortcut to the last tab? And while you're at it, make ctrl+tab (on mac) the "go to next tab" button as it is currently on both safari and chrome

frozeman:
So on mac super+number is the tab number, and ctrl + tab is next tab.

@evertonfraga can you please check why its not working for alex and also add the next tab ctrl + tab ?

evertonfraga:
Sure, checking out.

evertonfraga:
Oh, I get why it isn't working now. As I was refactoring, using auto-reload, it kept all the key bindings from previous delegations, when they were on `mistUI`. It's not working indeed @alexvandesande.

PR #1463, opened 2016-11-26
detect ropstein testnet

alexvandesande:
This adds the automatic detection of the Ropstein as “testnet” and not “privatenet”


frozeman:
Dont merge yet, as discussed with @alexvandesande he wants to make it more modular and display the name of the network in the `title=""` tag of the red button.

alexvandesande:
@frozeman I updated the code. Couldn't think of a way to include `web3.eth.getBlock(0,)` into the helper functions..

frozeman:
@alexvandesande where are you setting the name of the network to the title tag?
You can also pass the object to the `TemplateVar.set()`and then use it in the template as `network.type` or whatever its named.

PR #1477, opened 2016-11-29
ETH-02-cure-fixes #2

frozeman:
- ETH-02-001
- ETH-02-002
- ETH-02-003
- ETH-02-004


interface/client/lib/helpers/helperFunctions.js
```diff
@@ -114,7 +114,7 @@ Helpers.generateBreadcrumb = function (url) {
         return el === '';
     });
 
-    return new Spacebars.SafeString(_.flatten(['<span>' + filteredUrl.host + ' </span>', pathname]).join(' ▸ '));
+    return new Spacebars.SafeString(url.protocol +'//'+ _.flatten(['<span>' + filteredUrl.host + ' </span>', pathname]).join(' ▸ '));
```
evertonfraga: @frozeman here I'd recommend using the Blaze._escape the same way we're using on `host` and `pathname`.

frozeman: Will do, but as far as i understand whats put into the protocol is already stripped out. if invalid, it would be returned empty by webkit.

But we can certainly do it to increase security

PR #1479, opened 2016-11-29
updated electron to 1.3.12

frozeman:


luclu:
@frozeman what command did you use?
Doing `yarn upgrade electron@1.4.10` I get a way different yarn.lock file. [doc-link](https://yarnpkg.com/en/docs/cli/upgrade#toc-yarn-upgrade-package-version))

Also this will change the package.json entry to `"electron": "1.4.10"` (without the "="), which is unnecessary as "1.2.3" === "=1.2.3". (The "=" got introduces before "^" an was the only mean to lock a dependency)

If you want I can push a commit "fixing" those issues.

frozeman:
For me it broke the indexDB for the wallet for some reason and now it get errors where it can't open the index db of the wallet...

I guess i has to do with the update, maybe its more safe to update to the 1.3.12

Lets have a chat about that, before we merge.

luclu:
AFAIK indexDB issues are related to Chrome itself. 
Since electron `1.3.5` the only Chrome updates did happen in `1.4.0` and `1.4.6`, thus should we want to avoid the Chrome update for the next release the latest version without it is `1.3.6`.

PR #1483, opened 2016-11-30
Eth 02 fixes

frozeman:
Improves security in preloaders.

Fixes: 
- ETH-02-002
- ETH-03-001

Moves mist API to preloader include.
Fixes callback issue in Mist API.


evertonfraga:
From the code standpoint: 👍. Looked at every non-merge commit and they all seem fine.

I'll test against the ETH fixes, maybe monday morning.

modules/preloader/browser.js
```diff
@@ -28,6 +28,7 @@ process.on('loaded',function () {
 
 
 window.mist = mist();
+window.mistMode = remote.getGlobal('mode');
```
frozeman: the mistMode shouldn't be exposed to dapps in browser.js.
We also don't want to require remote her, for security reasons.

The issue should be fixed in the wallet.js

alexvandesande:
Security issues were tested.

PR #1453, opened 2016-11-23
[travis] revamp multi-platform build

luclu:
**Needs rebase on Taylor's merged PR**

fixes https://github.com/ethereum/mist/issues/1182

Currently splits the builds in (1: sudo trusty) linux + win and (2: osx) mac.

Outstanding TODO:
- [ ] container-based instead of full-virt
  - [ ] use `addons: apt:`, travis whitelist "PRs": 
    - [x] https://github.com/travis-ci/apt-package-whitelist/issues/3765 
    - [ ] https://github.com/travis-ci/apt-package-whitelist/issues/3766 
      - https://github.com/travis-ci/apt-source-whitelist/issues/339
    - [x] https://github.com/travis-ci/apt-package-whitelist/issues/3764
- [ ] osx build: certain steps take way too long, error out:
  - `copy-app-folder-files`: 1.48 min instead of 3.91 s (trusty)
  - `download-signatures`: errored after 1.52 min instead of 4.49 s (trusty)
```log
Selected platform: [ 'mac' ]
[14:01:05] Using gulpfile ~/build/ethereum/mist/gulpfile.js
[14:01:05] Starting 'mist'...
[14:01:05] Starting 'set-variables-mist'...
[14:01:05] Finished 'set-variables-mist' after 140 μs
[14:01:05] Starting 'download-signatures'...
[14:01:05] Starting 'clean:dist'...
[14:01:05] Finished 'clean:dist' after 28 ms
[14:01:05] Starting 'copy-app-source-files'...
[14:01:05] Starting 'copy-node-folder-files'...
[14:01:06] Finished 'copy-app-source-files' after 1.62 s
[14:01:06] Starting 'copy-app-folder-files'...
[14:03:35] Finished 'copy-app-folder-files' after 2.47 min
[14:03:35] Starting 'copy-build-folder-files'...
[14:03:35] 'download-signatures' errored after 2.5 min
[14:03:35] ParseError
    at stream.catch.then.data (/Users/travis/build/ethereum/mist/node_modules/got/index.js:114:14)
    at process._tickCallback (internal/process/next_tick.js:103:7)
[14:03:35] 'mist' errored after 2.5 min
[14:03:35] ParseError in plugin 'run-sequence(download-signatures)'
Message:
    Unexpected end of JSON input in "https://www.4byte.directory/api/v1/signatures/?page_size=20000&ordering=created_at": 
{"count":3337,"next":null,"previous":null,"results":[{"id":112,"created_at":"...
Details:
[...]

PR #1499, opened 2016-12-06
NSIS installer

tgerring:
Creating this PR for discussion and to track changes of updates to the Windows installer for Mist and integration into Travis. Working in conjunction with #1453 

cc/ @luclu


tgerring:
Current installer has 4 screens in total as outlined below (screeenshots via Wine on Mac):

screen shot 2016-12-06 at 17 36 27
Initial license screen (currently loading the file LICENSE)

screen shot 2016-12-06 at 17 36 38
Selecting the installation directory

screen shot 2016-12-06 at 17 36 51
Selecting the DATADIR, which is used to set shortcut links

screen shot 2016-12-06 at 17 37 07
Final installation screen with "show details"

cc/ @frozeman

frozeman:
How do you set the custom data folder in mist. And how the custom datadir for geth?

scripts/windows-installer.nsi
```diff
@@ -0,0 +1,135 @@
...
+    # Extract the zip file from TEMP to the user's selected installation directory
+    ZipDLL::extractALL "$TEMP\${RELEASEZIP}" "$FILEDIR"
+ 
+    # create the uninstaller
+    WriteUninstaller "$FILEDIR\uninstall.exe"
+ 
+    # create shortcuts with flags in the start menu programs directory
+    createDirectory "$SMPROGRAMS\${APPNAME}"
+    createShortCut "$SMPROGRAMS\${APPNAME}\${APPNAME} - Full Node.lnk" "$FILEDIR\win-unpacked\${APPNAME}.exe" "--fast --datadir=\"$DATADIR\""
+    createShortCut "$SMPROGRAMS\${APPNAME}\${APPNAME} - Light Client.lnk" "$FILEDIR\win-unpacked\${APPNAME}.exe" "--light --datadir=\"$DATADIR\""
```
frozeman: Please comment out the light client for now, as it is untested and not fully working with the wallet.

scripts/windows-installer.nsi
```diff
@@ -0,0 +1,135 @@
...
+    file "..\dist_mist\release\${RELEASEZIP}"
+    # Extract the zip file from TEMP to the user's selected installation directory
+    ZipDLL::extractALL "$TEMP\${RELEASEZIP}" "$FILEDIR"
+ 
+    # create the uninstaller
+    WriteUninstaller "$FILEDIR\uninstall.exe"
+ 
+    # create shortcuts with flags in the start menu programs directory
+    createDirectory "$SMPROGRAMS\${APPNAME}"
+    createShortCut "$SMPROGRAMS\${APPNAME}\${APPNAME} - Full Node.lnk" "$FILEDIR\win-unpacked\${APPNAME}.exe" "--fast --datadir=\"$DATADIR\""
```
frozeman: We should make clear that the data dir is the nodes data dir, and not mists data dir.
Mists data dir on mac is e.g. ~/Library/Application Support/Mist
Which contains the local storage for the browser, settings and the nodes executables

frozeman: Also you have to use `--node-datadir` if you want to change the data dir. fast is not necessary as we use that

tgerring: Should the installer have separate data directory pages for the node and Mist itself? Combined with installation location, this would total 3 directories to set

frozeman:
Currently we don't allow to set the Mist data dir manually, e.g. on windows its under %APPDATA%/Mist
And the Ethereum datadir is under %APPDATA%/Ethereum
I don't think we should change that or let the user even set that. If people want to use a custom one, thy can start the nodes themselves and are savvy enough to do so.

So i would only allow pople setting the installation folder for mist itself

tgerring:
Updated installer to allow user-settable location for Mist, Mist data files, node data files. The only effect of the latter two is to update the shortcut installed into the Start Menu. Additionally, we open a folder containing the Start Menu shortcuts on successful installation. If the user wants to modify the settings after installation, they only need to right-click the shortcut and modify the parameters it passes.

A few TODOs:
* Add file size estimates
* Decide how to handle multiple architecture: bundle both into a single large installer, or generate separate installers
* Pass version information from gulp build script instead of hardcoding values

frozeman:
So its still asking now for 2 folders? Which are the default ones set there?

tgerring:
Based on the current commit, the user is prompted for 3 locations, all of which are defaulted to the value the Mist team has outlined in the PR and/or the README:

* Mist program files
* Mist data files (--DATADIR)
* Node data files (--NODE-DATADIR)

At this point, I feel that more input is needed from the team before the installer can be further revised

frozeman:
As of now we don't support setting the user data manually using `--datadir`, we only pass the `--node-datadir` as `--datadir` to the node.

As i said i don't know how useful it is, especially for windows user.

As i don't feel comfortable adding stetting the user data path manually right now, as i don't see the use for that right now.

Could you comment setting these datadirs manually for now? We might reactivate that later, when we revisit this idea. But for now we only want to set the installation directory for Mist.

We also would like to publish a new mist version today, if you can make the installer work with only one folder we could probably include it in this release, after testing.q

tgerring:
# Installation step preview:

screen shot 2016-12-13 at 09 59 22
screen shot 2016-12-13 at 09 59 30
screen shot 2016-12-13 at 09 59 42
screen shot 2016-12-13 at 09 59 49
screen shot 2016-12-13 at 10 00 09
screen shot 2016-12-13 at 10 00 33

# Uninstallation step preview:
screen shot 2016-12-13 at 10 03 18
screen shot 2016-12-13 at 10 03 44
screen shot 2016-12-13 at 10 03 50
screen shot 2016-12-13 at 10 03 59
screen shot 2016-12-13 at 10 04 13

frozeman:
Looks great, tho i need to ask you to comment out the Mist App Datadir selection, as thats not possible right now and we wont add it for some time (we need to think about that useful it actually is cross platform)

The popping up of the data folders on uninstall is great, and i recommend using the default Mist data folder for that, as the user can't select any.

frozeman:
Hi Taylor. So i tried the installer and it seems to work perfectly.
A few notes, tho:

- The estimated space is half of that what it on the end needs. Estimated 125MB, used on the end 250MB.
- The Mist-win32/64....zip is not removed from the `...local/Temp` folder
- The bundled interface dapp is broken, i guess you forgot to call `meteor npm install --save babel-runtime` from inside the mist/interface folder. @luclu we need to add that to the readme and maybe to the gulp script?
- the uninstall script doesn't open the ethereum folder for key backup. Would be nice to add that back.

What happens when i install multiple times mist, without uninstalling the previous one?

luclu:
> The bundled interface dapp is broken, i guess you forgot to call meteor npm install --save babel-runtime from inside the mist/interface folder. @luclu we need to add that to the readme and maybe to the gulp script?

@frozeman done in PR: https://github.com/ethereum/mist/pull/1546

frozeman:
@tgerring we had a discussion about the current issues and want to change the following:

- Create Desktop icon and start menu entry, just for "Mist"
- Remove testnet startmenu entry (as this needs to be started from within mist, to show correctly in the menu and allow future customisations)
- Remove zips from temp after install
- we added a `cd interface && yarn` to the post install in the package.json. Please run this manually to make test builds work properly on your side.
- Is it possible to fix the incorrect size estimation? If not thats just a  beauty issue

tgerring:
@frozeman It would be better if we could discuss some of these issues directly, as things like the size issue are a consequence of how things are packaged before NSIS

PR #1500, opened 2016-12-06
Restart ethblocks

alexvandesande:
It’s a small fix that solves the issue of the sidebar not updating the blocks because ethBlocks is not keeping up with web3.eth.getBlocks.

On `nodeInfo.js`, whenever `syncing == true`, it will execute a `web3.reset(true);` which stops all listeners. But when syncing stops again it does a MistInit but not a EthBlocks.init(). So I’m putting the EthBlocks.init() inside MistInit so it will restart it. It’s not called elsewhere.

### DIFF
--- a/interface/client/appStart.js
+++ b/interface/client/appStart.js
@@ -12,6 +12,8 @@ The init function of Mist
 mistInit = function(){
     console.info('Initialise Mist Interface');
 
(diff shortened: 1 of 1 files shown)


evertonfraga:
@alexvandesande it's working amazingly :)

👍

PR #1501, opened 2016-12-07
Adding dapp tests

evertonfraga:
Added tests in the context of the dapps. Also, some checks over `web3` and `mist` apis, regarding which keys should and shouldn't be allowed on the dapp scope.

Added new menu item `Develop > Run tests anonymously`, with the according test suite.


frozeman:
Maybe you want to fix the code climate issues also

PR #1503, opened 2016-12-09
Mist API tests

evertonfraga:
- `mist.platform`
- `mist.menu.select(id)`


modules/preloader/include/mistAPI.js
```diff
@@ -159,20 +157,23 @@ module.exports = () => {
                 const filteredId = prefix + filterId(id);
```
frozeman: Please also add the comments, like the other functions have not above.

modules/preloader/include/mistAPI.js
```diff
@@ -159,20 +157,23 @@ module.exports = () => {
...
                 queue.push({ action: 'selectMenu', id: filteredId });
 
-                for (var e in this.entries) {
+                for (let e in this.entries) {
                     if ({}.hasOwnProperty.call(this.entries, e)) {
                         this.entries[e].selected = (e === filteredId);
```
frozeman: The other ones need to set `selected = false`, you only seem to set the new one to true, which means multiple ones have true now.

modules/preloader/include/mistAPI.js
```diff
@@ -159,20 +157,23 @@ module.exports = () => {
...
                     }
                 }
+                executeCallback(null, filteredId);
```
frozeman: Not sure, if calling `select(id)`, should also call the callback for that menu entry..
this api is mainly for updating the menu in mist, not for calling actions in the developers dapp. For this he will have to program his own logic.

The idea would be also that the mist api is not available sometimes, so you wouldn't depend core logic of your app on it. You would want to do it like this:

```js
'click .mySuperButton': function(){

    MyApp.goToPath('/super');

    // also select entry, if mist is available
    if(window.mist) {
        mist.select('superButton');
    }
}
```

frozeman:
Also where is the logic on the backend for selecting the menu entry? I couldn't find another PR and its not in devleop

modules/preloader/include/mistAPI.js
evertonfraga: @frozeman actually this traverses through all entries, where `e === filteredId` returns either `true` or `false`, depending on the current element.

evertonfraga:
@frozeman this logic is here: https://github.com/ethereum/mist/blob/develop/interface/client/mistAPIBackend.js#L126-L130

modules/preloader/include/mistAPI.js
evertonfraga: yeah, that makes sense indeed. updated.

evertonfraga:
I'll first add the 100 menu entries limit before merge. Thanks for the changes

PR #1520, opened 2016-12-16
Update geth to 1.5.5

evertonfraga:
Testing checklist
- [x] Checksums
- [x] Mac
- [x] Win32
- [x] Win64
- [x] Linux32
- [x] Linux64

PR #1526, opened 2016-12-16
Update geth to 1.5.5 (#1520) (#1522)

gentlejack:
* updated solc again

* updated solc to 0.4.6

* Update geth to 1.5.5 (#1520)

* Update geth to 1.5.5

* corrected windows zip's internal paths

* changed sanity check

* made node starting better

* also log path from which to fetch

* show fetching origin URL

* fix ESLint errors

* removed -stable

### DIFF
--- a/clientBinaries.json
+++ b/clientBinaries.json
@@ -2,36 +2,36 @@
 {
     "clients": {
         "Geth": {
-            "version": "1.5.3",
+            "version": "1.5.5",
             "platforms": {
                 "linux": {
                     "x64": {
                         "download": {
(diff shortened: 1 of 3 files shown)


evertonfraga:
hi @gentlejack. what is the reason of this PR?

thanks

luclu:
Basically because our branching still can be considered as 'freestyle' and git can't cope with our artistic freedom:

We might wan't to do a two-way merge to rectify it:
1. `master` into `develop`
2. `develop` into `master`

branching

frozeman:
Closing this as its already up to date in develop

PR #1543, opened 2016-12-22
Adjust formatter for addPeer

genecyber:
enode://... doesn't convert to decimal very easily 😀

### DIFF
--- a/modules/web3Admin.js
+++ b/modules/web3Admin.js
@@ -13,7 +13,7 @@ module.exports = {
             property: 'admin',
             methods:
             [
-                insertMethod('addPeer', 'admin_addPeer', 1, [web3._extend.utils.fromDecimal], web3._extend.formatters.formatOutputBool),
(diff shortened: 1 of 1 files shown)


luclu:
@genecyber could you please push a dummy commit to get travis (our continuous integration service) running? 
There seems to be some travis-internal problem and this PR did never get into the build queue.
We will be able to merge it as soon as the required check did finish.

Thanks!

frozeman:
It never seems to build

genecyber:
Should I still try and do a other commit to this PR? Or is this a bigger problem?

luclu:
@genecyber yes please, we never encounter this issue with travis, so our hope is that it will build if you push another commit adding or removing just a empty line in some file for example.

If this does not help I will contact travis support to solve this issue on their side.

luclu:
I asked travis to retrigger the stalled build.

luclu:
Travis support is unable to restart, so I recreated the commit: https://github.com/ethereum/mist/pull/1599

PR #1547, opened 2016-12-22
Light client 

alexvandesande:
Second PR, focusing on light client issues only. 

Review but do not merge before more group discussion.


alexvandesande:
goes better after https://github.com/ethereum/mist/pull/1545/files

modules/ethereumNode.js
```diff
@@ -130,7 +139,7 @@ class EthereumNode extends EventEmitter {
                 log.info(`Network: ${this.defaultNetwork}`);
```
frozeman: Would be good to add: `  log.info(`Mode: ${this.defaultMode}`);`

modules/ethereumNode.js
```diff
@@ -140,20 +149,20 @@ class EthereumNode extends EventEmitter {
...
-            log.info('Restart node', newType, newNetwork);
+            log.info('Restart node..', newType, newNetwork, newMode);
 
             return this.stop()
                 .then(() => {
                     Windows.loading.show();
                 })
                 .then(() => {
-                    return this._start(newType || this.type, newNetwork || this.network);
+                    return this._start(newType || this.type, newNetwork || this.network, newMode || this.nodeMode);
```
frozeman: Whats this `this.nodeMode` ? i never see this variable created,shouldnt it be `this.mode` or something?

alexvandesande:
@frozeman addressed both issues

modules/ethereumNode.js
```diff
@@ -162,7 +163,7 @@ class EthereumNode extends EventEmitter {
                     Windows.loading.show();
                 })
                 .then(() => {
-                    return this._start(newType || this.type, newNetwork || this.network, newMode || this.nodeMode);
+                    return this._start(newType || this.type(), newNetwork || this.network(), newMode || this.mode());
```
frozeman: thats not functions, this are get variables.
So just `this.mode`

alexvandesande: I've tested and `this.mode()` was returning a value and `this.mode` wasn't. Are you sure this isn't being made into a function by https://github.com/ethereum/mist/pull/1547/files#diff-409860c680be3333b2089d577f193cf5R65 ?

Will test more this part

evertonfraga:
- When the light client is far behind and I'm on the app screen (not splash), mist hits high CPU usage until the node catches up. That may cause unresponsiveness on slower machines (mine is a quad-core 2.2 GHz)
- Mist slowness was on the main thread, making the actual interface sluggish
- When I'm on light client mode and disconnect from internet and connect again, it didn't seem to automatically reconnect – or there weren't many peers available on my tests.

alexvandesande:
superseded by https://github.com/ethereum/mist/pull/3052

Issue #1066, opened 2016-08-04
Show Error message instead of Blank page on DNS failure


jorisbontje:
When I enter an URL with a domain that doesn't exist (like www.mistgivesblankpage.com), Mist shows a blank page (see screenshot beelow).

![screenshot 2016-08-04 13 31 28](https://cloud.githubusercontent.com/assets/620959/17400414/ff17f6c2-5a47-11e6-9a4b-8d34e2d9ac74.png)

Instead I would expect to see some kind of DNS error page like how Chrome does:

```
This site can’t be reached

www.mistgivesblankpage.com’s server DNS address could not be found.
Search Google for misty gives blank page
ERR_NAME_NOT_RESOLVED
```
#### System information

```
Version: `0.8.1`
OS & Version: osx
Node type: `geth(default)`
```

morelazers:
I think this might be related to a problem I am having, even though the page does exist (I can visit it in Chrome for example).

Is there any news on this issue?

frozeman:
We currently don't have error page for that case, but it certainly makes sense to add one!

frozeman:
Here is the PR https://github.com/ethereum/mist/pull/1559

PR #1559, opened 2017-01-06
Error pages

frozeman:
Adds error pages for not found URLs fixes https://github.com/ethereum/mist/issues/1066#issuecomment-270661218

url not found
screen shot 2017-01-06 at 11 13 22

crash
screen shot 2017-01-06 at 11 25 00

We can add more error pages later on


alexvandesande:
How can I force an error page?

Issue #1569, opened 2017-01-10
Edge-case offline-startup error


luclu:
Currently Mist will error-out on the splash screen if (i) no internet connection is available and (ii) the downloaded version of geth differs from the one advertised in the bundled clientBinaries.json.

From https://github.com/ethereum/mist/pull/1534:
>There seems to be a case where this fix does not suffice.
>
>If the default config - shipped with ethereum-client-binaries - differs from the current version (e.g. of geth), I think it will try to download the older version (see clientBinaryManager.js#L181-L184). Also we can't keep the default config up-to-date on shipped installations.

#### Proposals

1. Disable version check of already downloaded node if clientBinaries.json can't be received from github
2. Store version of currently downloaded version in local DB

I would prefer (1).

evertonfraga:
afaik, Mist also saves the clientBinaries.json on `Library/Mist`. You could
use that version to check against the geth version

PR #1571, opened 2017-01-10
fix offline-startup edge-case

luclu:
fix https://github.com/ethereum/mist/issues/1569
Prefer the downloaded `clientBinaries.json` over the bundled one if offline.
This will prevent issues if the downloaded node's version isn't on parity with the bundled one.

### DIFF
--- a/modules/clientBinaryManager.js
+++ b/modules/clientBinaryManager.js
@@ -180,7 +180,9 @@ class Manager extends EventEmitter {
         .then((localConfig) => {
             if (!localConfig) {
                 log.info('No config for the ClientBinaryManager could be loaded, using local clientBinaries.json.');
-                localConfig = require('../clientBinaries.json');
+
+                const localConfigPath = path.join(Settings.userDataPath, 'clientBinaries.json');
+                localConfig = (fs.existsSync(localConfigPath)) ? require(localConfigPath) : require('../clientBinaries.json');  // eslint-disable-line no-param-reassign, global-require, import/no-dynamic-require, import/no-unresolved
             }
 
             // scan for node


modules/clientBinaryManager.js
```diff
@@ -180,7 +180,9 @@ class Manager extends EventEmitter {
         .then((localConfig) => {
             if (!localConfig) {
                 log.info('No config for the ClientBinaryManager could be loaded, using local clientBinaries.json.');
-                localConfig = require('../clientBinaries.json');
+
+                const localConfigPath = path.join(Settings.userDataPath, 'clientBinaries.json');
```
frozeman: I dont understand what is the fix. On line 87 it already tries to load the local config. Then later it will replace it with the remote loaded one, if one came if not it keeps the locally loaded one. On this line it basically already checked local and remote ones and therefore loads the one from the mist folder.

luclu: The statement in L#87 can only be reached if the computer is online. Otherwise it will simply return to the next promise-chain block. The skipped validation doesn't seem to be necessary as this did happen before as the `clientBinaries.json` was created.

frozeman: As far as i understand this block:
```js
// FETCH
 return got(BINARY_URL, {
             timeout: 3000,		             timeout: 3000,
             json: true,		             json: true,
         })		         })
// GET RESULT OR ERROR
         .then((res) => {		         .then((res) => {
             if (!res || _.isEmpty(res.body)) {		             if (!res || _.isEmpty(res.body)) {
                 throw new Error('Invalid fetch result');		                 throw new Error('Invalid fetch result');
             } else {		             } else {
                 return res.body;		                 return res.body;
             }		             }
         })		         })
// CATCH ERRORS from fetching, on or offline
         .catch((err) => {		         .catch((err) => {
             log.warn('Error fetching client binaries config from repo', err);		             log.warn('Error fetching client binaries config from repo', err);
         })		         })
// CONTINUE here, even if error was above thrown.
         .then((latestConfig) => {		         .then((latestConfig) => {
```

Please test this that a offline state will certainly not go into this then block after the catch and the duplication of the code is really necessary

luclu: Yes, that's right. The next line will simply return and jump to the next block. As the `localConfig` is empty it will use my new code to fetch the already downloaded. If that one is not available (first run of Mist is offline) it uses the bundled. Those configs are both validated and so this plays nice with the omiting of the validation block.

PR #1574, opened 2017-01-11
update geth to 1.5.6

luclu:
Tested download on:
- [x] macos
- [ ] win32
- [ ] win64
- [x] linux32
- [x] linux64

**Update: geth 1.5.6 will be skipped as in favour of 1.5.7 as the current release is incompatible with Mist**

### DIFF
--- a/clientBinaries.json
+++ b/clientBinaries.json
@@ -2,36 +2,36 @@
 {
     "clients": {
         "Geth": {
-            "version": "1.5.5",
+            "version": "1.5.6",
             "platforms": {
                 "linux": {
                     "x64": {
                         "download": {
-                            "url": "https://gethstore.blob.core.windows.net/builds/geth-linux-amd64-1.5.5-ff07d548.tar.gz",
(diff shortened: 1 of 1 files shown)


karalabe:
On the pub channel someone wrote:

> Greg @ethminer1_twitter 01:52
Hi all, new to here ....
wasnt sure how/where to report, but wanted to make sure that there's awareness that the new geth node (v1.5.6) causes ethereum wallet to stop working. Shutting down the new client and letting the wallet use the internal client still works fine. Where best to voice that?

Could you verify whether everything at least seems to work with the update before pushing it out? Thanks

luclu:
Thanks Péter, I'm still testing, but so far it LGTM

luclu:
Incompatibility issue:
> Sending gives error 'invalid argument 0: empty hex string' https://github.com/ethereum/meteor-dapp-wallet/issues/334

luclu:
Update: geth 1.5.6 will be skipped as in favour of 1.5.7 as the current release is incompatible with Mist

Issue #1570, opened 2017-01-10
Rectify filenames when using travis to publish artifacts


luclu:
Electron-builder's built-in artifact uploader (e.g. github releases) will use the automatically generated filenames for the binaries and omit our post-hook renaming.

Investigate if:
1. those names can be updated internally, which would make the post-hook obsolete
2. a gulp compatible uploader should be used instead of electron-builder's internal

***Note:*** electron-builder doesn't support uploading to azure, if this was planned to be used [gulp-azure-storage](https://www.npmjs.com/package/gulp-azure-storage) could be leveraged instead (2).

luclu:
#### Update 

(1) looks like it requires a PR to electron-builder
(2) seems feasible using [gulp-github-release](https://github.com/Aluxian/gulp-github-release), though need to check its behaviour (i.e. if it will create tags on its own); + switching to [gulp-azure-storage](https://www.npmjs.com/package/gulp-azure-storage) will be easy
3. use non-final binary names from electron-builder to upload and use github's API to rectify the names

luclu:
Route (2)'s progress on https://github.com/ethereum/mist/pull/1578

PR #1578, opened 2017-01-11
[gulp] new task 'upload-binaries' (to github)

luclu:
This PR allows Travis to upload the generated binaries (`dist-${type}/release`) to github.

It will only do so if:
1. building on `master` branch (after a PR got merged from `develop` or direct commit)
2. a git tag `v{version}` exists
3. a release in **draft** mode coupled with the tag from (2) exists
4. the draft does not contain any files with the same name that the generated binaries have

This solves:
- direct CI upload
- can't use electron-builder's internal uploader because:
  - naming problem using electron-builder to upload (#1570) 
  - using custom NSIS script (https://github.com/ethereum/mist/pull/1499) to create windows installer a posteriori 

Also it will allow for simple integration of upcoming azure uploader.

I extensively tested this PR on my fork of the repo.


frozeman:
So the tag needs to exist before while the release is still in draft?
I would suggest that the release creates the tag on publish how we do it now

luclu:
Eventually we decided to create the tag a prior - Ev reminded me that to sign tags those must be created locally pushed. Did I understand this correct, @evertonfraga?

Otherwise letting Travis create the tag + draft if the version gets bumped on `master` sounds resonable. (Although the upload should also work if the draft already exists/was created by us beforehand.)

alexvandesande:
Thanks for this @luclu please review the conflicts

.travis.yml
```diff
@@ -10,6 +10,7 @@ cache:
 branches:
   only:
     - develop
```
frozeman: Why we upload on develop?

gulpfile.js
```diff
@@ -347,14 +356,54 @@ gulp.task('release-dist', ['build-dist'], (done) => {
     done();
 });
 
+gulp.task('upload-binaries', () => {
+    // token must be set using travis' ENVs
+    const GITHUB_TOKEN = process.env.GITHUB_TOKEN;
+
+    // query github releases
+    return got(`https://api.github.com/repos/ethereum/mist/releases?access_token=${GITHUB_TOKEN}`, {
```
frozeman: Is that token secret?

frozeman: Ah i see it comes from travis

.travis.yml
luclu: The `branch -> only` array includes all branches that should trigger a travis build. Thus we want `develop` and now `master` for the upload.

gulpfile.js
luclu: Correct.

PR #1585, opened 2017-01-13
 Adjustments to confirmation popup scroll-area

luclu:
(This has been bugging me for some time now)

Here is the **current state** of the confirmation window:
1. the function parameters are shown before the fee information
2. no indication of scrollable area: it is not obvious that there is further content below the parameters block, the bottom of the scroll area has a small often invisible feather (no feather at the top).

---
**Current state:** scrolled to top (just opened)

bildschirmfoto 2017-01-13 um 14 44 46

**Current state:** scrolled to bottom

bildschirmfoto 2017-01-13 um 14 45 01

---

**Proposed changes:**
1. switch order of (a) function params and (b) fee info
2. subtle changes to feather and window size to cut the line at a position that makes the feather visible and more obviously indicates the scroll area

**Proposed change:** scrolled to top (just opened)

bildschirmfoto 2017-01-13 um 15 02 36

**Proposed change:** scrolled to top (just opened)

bildschirmfoto 2017-01-13 um 15 02 52


evertonfraga:
I recall that @alexvandesande changed to that setup in favor of having an easy way for users to double-check the params.

alexvandesande:
That has been bugging me since the change too. I wonder if maybe by using a fee slider we can simplify the whole gas price/fee interface and make it smaller? I wonder if there's a way to include both price and gas in one simple button.

PR #1586, opened 2017-01-13
[mistAPI] add 'mist.solidity.version'

luclu:
I still wonder if we should add `mist.compiler.solidity.version` instead?


luclu:
Thank Ev! 👍

evertonfraga:
👍 make sure to run it when making changes to it.

PR #1587, opened 2017-01-13
fix mention-bot

luclu:
...as evaluation logic got updated. 
Now necessates `"actions": ["labeled"]` if `"withLabel": "please review"` was to be used.


evertonfraga:
@luclu would that work if one puts the label **after** creating the PR?

luclu:
Yes, as a default mention-bot will listen to the "opened" event/action and create the comment. On the other hand the "with label" parameter only works with the "labeled" acition.

PR #1608, opened 2017-01-23
Adding a develop menu option to open Remix IDE

evertonfraga:
I'd suggest we use a shorter name on `<title>` tag, so it displays nicer on Mist interface, and a more straightforward icon would help as well :)  

@yann300 @chriseth I could raise a PR on remix repo for that, just tell me the directions (image file and new title)

image

image


modules/preloader/mistUI.js
```diff
@@ -131,6 +131,17 @@ ipcRenderer.on('uiAction_runTests', (e, type) => {
     }
 });
 
+// Open Browser Solidity
+ipcRenderer.on('uiAction_openBrowserSolidity', () => {
+    Tabs.upsert('browserSol', {
+        position: Tabs.find().count(),
+        name: 'Remix',
+        redirect: 'https://ethereum.github.io/browser-solidity/',
```
chriseth: Please use `https://remix.ethereum.org`

modules/menuItems.js
```diff
@@ -357,9 +358,17 @@ let menuTempl = function (webviews) {
                 log = 'Couldn\'t load log file.';
             }
         },
-    },
-    ];
+    }];
 
+    if (Settings.uiMode === 'mist') {
+        devToolsMenu.push({
+            label: i18n.t('mist.applicationMenu.develop.browsersol'),
```
chriseth: I think it is better to use `remix` throughout instead of `browsersol` or `BrowserSolidity` (below).

chriseth:
@evertonfraga good idea! Please change the title to just "Remix" in github.com/ethereum/browser-solidity. We could also use "Remix - Solidity Realtime...".
What do you mean by "more straightforward icon"?

frozeman:
Ok please rename to "Open Remix IDE" on the menu.

Then please open in a `windows.createPopupWindow`
This should then have web3 and all accounts by default.

yann300:
cool! I sent you logos in the gitter channel that you could use.

evertonfraga:
- [x] Change url to `https://remix.ethereum.org`
- [x] Open Remix in a popup, with access to all accounts
- [x] Use `Remix` naming instead of `browserSol` throughout code

Opening (the recoding is cropped, Remix opens centered): 

remix mov


Accounts: 

remix-2 mov

chriseth:
That looks nice! Is it obvious to the user that this remix instance is tied to mist?

frozeman:
Yes it is as he opens it from the Mist menu. Tho we have to improve the visibility of the current running node (type) and maybe also improve that in Remix. E.g. @chriseth @yann300 

For example highlighting the current provider used (JS, web3, etc) somewhere  always visible in the Remix interface.

frozeman:
@evertonfraga it would be good if we set the title of the popup to "Remix IDE"

frozeman:
@chriseth i would also prefer a dark theme for the Remix editor. this would immediately make the whole interface cleaner looking as the editor pane will be clearly separated from the debug panel

evertonfraga:
@frozenan There'll be a PR for Remix repo, where i'll change this title,
and try to fix the `window.prompt` compatibility issue.

frozeman:
@evertonfraga please push another empty commit, something is wrong with travis

PR #1609, opened 2017-01-25
Update app.nl.i18n.json

h3ll0fr13nd:
Dutch translation


frozeman:
I would always choose informal

PR #1637, opened 2017-02-02
Add gulp task 'update-nodes'

luclu:
This gulp-task will:
1. query the latest geth version from its github releases
1. compare its version to 'clientBinaries.json' and if newer will
1. query github for the commit hash and azure for the md5 checksums
1. update 'clientBinaries.json' accordingly

Its execution takes:
- ~100ms if no update available
- 5-10s if update is available


evertonfraga:
👍 👍

PR #1641, opened 2017-02-03
Persisting main window bounds

evertonfraga:
Once this is tested on windows and linux, can be merged.

It reopens the window with a minimum width and height, to avoid mistakes/confusion on the user side. Also, the window always spawns inside the screen bounds.


obscuren:
Yay, awesome!

PR #1645, opened 2017-02-05
Add low cpu priority for geth on UNIX-like systems (fixes #1620)

### LINKED ISSUE #1620: Ethereum Wallet should start geth with low priority
#### System information
``````
Version: `0.8.8`
OS & Version: debian testing
Node type: `geth(default)`
``````
During sync geth runs in background with standard priority and consumes a lot of CPU power for a long time. The usability of my old laptop (IBM T60p) goes down. If geth runs with lower priority (renice) the usability is much better.

dagnyswaggart:
Hey I'm brand new to the ethereum community and I figured I'd start by getting the wallet up and making a small contribution and learn from the feedback from more experienced contributors. I found #1620 and figured I'd give it a shot.

on 'linux', 'darwin', 'freebsd', 'sunos' platforms it will execute renice after geth is run and set it to a given priority (currently 19, could possibly be made configurable or changed?)

I don't know if I did this right, but if I didn't there's my opportunity to learn how :)


evertonfraga:
Such a clever idea @dagnyswaggart. what's the rationale behind it, Give Mist interface less chance to slow responsiveness when mining? What use case did you envision?

dagnyswaggart:
@evertonfraga the use case of satisfying @schwitin in #1620 I guess lol. Just tried to pick an easy issue and fix it. Renice is the tool for changing cpu priority - I googled how to do it in node and there's no built in way. A google groups post said to spawn your process with renice. I would have had to rewrite the code that runs the node but I checked if renice supports pid - it does. then I just set it to the lowest priority possible - it's possible a reviewer will want me to set it to something else like 15 or decide that the cpu priority shoukdn't be changed at all, but I figured I'd do my best to resolve an open issue and get my foot in the door in the community - that was my ultimate goal :)

frozeman:
What is `proc` where does it come from? Which pid is it?
And why setting it to lower, if `miner.start(1)` already uses only one CPU when minion on the test chain?

kenshyx:
On windows you can use `wmic` to set cpu priority.

frozeman:
I am not sure if we should merge that.
A lower (lowest!) default CPU prio, could significantly change the users ability to sync. While this might help on lower end machines, it makes the overall usage worse, and even dapps which call geth all the time will be slower.

So i am not convinced by this change at all.

evertonfraga:
We'll re-evaluate it after the refactor we're making. Thanks for the contribution ;)

PR #1680, opened 2017-02-27
Last days of dechunker

frozeman:
This removed the DECHUNKER and adds oboe.js as a better solution.


bas-vk:
Just wanted to mention that when I created the poc with oboe my goal was to show you guys that it's possible to parse json from an endless stream without the need for delimiters or something equivalent. Please make sure that oboe and its underlying parser clarinet are production ready and suitable for Ethereum use. Clarinet has a few issues that have been open for quite some time and only received 1 commit in almost 2 years. Now that you know where to look for it might be interesting to spend some time to search for alternatives.

Some things to note:
- clarinet supports data of [MAX_BUFFER_LENGTH](https://github.com/dscape/clarinet/blob/master/clarinet.js#L14) (64k). That might not be enough in come cases. E.g. `eth_getCode` for really large contracts (note that data is hex encoded and doubled in size).
- https://github.com/dscape/clarinet/issues/40 can increase Mists bundle size significantly until a new release comes out.

frozeman:
Seems to work:
screen shot 2017-02-27 at 14 03 13

PR #1689, opened 2017-03-01
More spectron ETH-01 tests.

evertonfraga:


tests/mist/basic.test.js
```diff
@@ -151,15 +151,15 @@ test['"data:" protocol should be disallowed on browser bar'] = function* () { //
     browserBarText.should.eql('http://localhost:8080'); // checks that hasn't changed displayed URL
 };
 
-// test['"file:///" protocol should be disallowed'] = function* () { // ETH-01-002
-//     const client = this.client;
-//     const filePath = 'file://' + path.join(__dirname, '..', 'fixtures', 'index.html');
+test['"file:///" protocol should be disallowed'] = function* () { // ETH-01-002
```
luclu: Does the comment string/this test actually refert to `ETH-01-012` instead of `ETH-01-002`?

luclu: Ah, I think the vuln-code comments are invalid. They seem to be copied and not adjusted, is this correct?

evertonfraga: you're right. that should be 012. 002 is about omission of scheme and it's been dealt with on previous tests.

PR #1774, opened 2017-03-20
Update README.md

Projjol:
Add yarn as a dependency in the README file

### DIFF
--- a/README.md
+++ b/README.md
@@ -41,6 +41,7 @@ To run mist in development you need [Node.js NPM](https://nodejs.org) and [Meteo
     $ curl https://install.meteor.com/ | sh
     $ npm install -g electron@1.3.13
     $ npm install -g gulp
+    $ npm install -g yarn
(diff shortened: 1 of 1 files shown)


luclu:
Thank you for the PR, @Projjol!

The yarn community does advice to refrain from installing yarn via npm (see [here](https://yarnpkg.com/en/docs/install#alternatives-tab)):
>  Note: Installation via npm is generally not recommended. npm is non-deterministic, packages are not signed, and npm does not perform any integrity checks other than a basic SHA1 hash, which is a security risk when installing system-wide apps.
>
> For these reasons, it is highly recommended that you install Yarn through the installation method best suited to your operating system.

Although yarn is listed as a dependency in the [Readme](https://github.com/ethereum/mist#dependencies), it is not very clear.

I will take a look at it today and rephrase the paragraph. Thanks again for pointing this out!

PR #1775, opened 2017-03-20
show backup hint when createing accounts, demanding min 8 characters

frozeman:
as the name says


interface/i18n/mist.en.i18n.json
```diff
@@ -182,6 +182,7 @@
                 "enterPassword": "Enter password",
                 "repeatPassword": "Repeat password",
                 "creating": "Generating account...",
+                "backupHint": "Make sure you backup you keyfiles AND Password!\n\nYou can find your keyfiles folder using the main menu -> Accounts -> Backup -> Accounts. Keep a copy of the \"keystore\" folder where you can't loose it!",
```
luclu: typo `you` -> `your`.

PR #2092, opened 2017-03-31
[security] fix #1359, improve txData validation

### LINKED ISSUE #1359: Insufficient input validation sendTransaction
#### System information

```
Version: `0.8.1-0.8.7`
OS & Version: all
Node type: `all`
```
#### Description

The the following payload tries to inject javascript into the elevated-privilege signature popup, by converting the payload to a RegExp:

```
var payload = 'document.body.innerHTML=\'Your accounts:\' +web3.personal.listAccounts.join(\'<li>\');return false;'
var xss = '</pre></div><button onclick="'+payload+'">AAA</button></div>'
var add = "0x0000000000000000000000000000000000000000";
web3.eth.sendTransaction({ from : add, to : add, data: new RegExp(xss)});
```

This attack fails, with the following stack-trace in the console: 

```
  message: 'json: cannot unmarshal object into Go value of type string' }
[2016-10-28 09:24:00.306] [INFO] (ui: popup) - Exception in template helper: TypeError: this.data.replace is not a function
    at Object.formattedData (file:///data/tools/Mist-linux64-0-8-6/linux/resources/app.asar/interface/a9937a6cd93a5d79cd8de8f368032a8586fe93f1.js:283:3624)
    at file:///data/tools/Mist-linux64-0-8-6/linux/resources/app.asar/interface/a9937a6cd93a5d79cd8de8f368032a8586fe93f1.js:88:30879
...
```

Interestingly, the reason this attack fails is this line: https://github.com/ethereum/mist/blob/c3ae7606ed441f2aad11ed4542fb7ced47ede00e/interface/client/templates/popupWindows/sendTransactionConfirmation.js#L240 

```
    'formattedData': function(){
        return (TemplateVar.get('toIsContract'))
            ? this.data.replace(/([0]{2,})/g,'<span class="zero">$1</span>').replace(/(0x[a-f0-9]{8})/i,'<span class="function">$1</span>')
            : this.data.replace(/([0]{2,})/g,'<span class="zero">$1</span>');
    },
```

Basically, what happens is that the UI rendering tries to invoke the `replace` on a `RegExp` object, and fails. 

The actual input validation is a bit flawed, https://github.com/ethereum/mist/blob/c3ae7606ed441f2aad11ed4542fb7ced47ede00e/modules/ipc/methods/eth_sendTransaction.js#L40:

```
                    if(_.isString(val)) {
                        if (val.match(/[^0-9a-fx]/igm)) {
                            throw this.ERRORS.INVALID_PAYLOAD;
                        }
                    }
```

Since the input is a `RegExp`, the `isString` returns false, and the intended validation that the input is hex-digits is bypassed. 

**This is very fragile**, since a change to either the template used or the template-formatting function would open the signing window for XSS-attacks. 
#### Suggested remediation
- Ensure that all input elements are expected type. Afaict, either `number` or `string`. 

Example

```
x=/test/;
if(typeof(x)=='number'){ val=x;}
else{val=String(x)} 
```

luclu:
fix #1359


modules/ipc/methods/eth_sendTransaction.js
```diff
@@ -36,14 +36,16 @@ module.exports = class extends BaseProcessor {
             try {
                 _.each(payload.params[0], (val, key) => {
                     // if doesn't have hex then leave
-                    if (_.isString(val)) {
+                    if (_.isString(val) || _.isNumber(val)) {
```
holiman: Won't this throw when you try to do `toLowerCase()` on the number a few lines down?

modules/ipc/methods/eth_sendTransaction.js
```diff
@@ -36,14 +36,16 @@ module.exports = class extends BaseProcessor {
...
 
                         // make sure all data is lowercase and has 0x
                         if (val) val = `0x${val.toLowerCase().replace('0x', '')}`;
```
holiman: This isn't part of the PR, I know... but doing `toLowerCase()` here removes the checksum (if any), and could be problematic later on if you want to add checksum verifications..

modules/ipc/methods/eth_sendTransaction.js
```diff
@@ -36,14 +36,16 @@ module.exports = class extends BaseProcessor {
...
 
                         if (val.match(/[^0-9a-fx]/igm)) {
```
holiman: Why not be even stricter - this would allow arbitrary mix of hex-digits and `x`. Why not require

* Starts with `0x`
* Then must be `0-9a-fA-F`

Something like `/0x[0-9a-fA-F]/`

modules/ipc/methods/eth_sendTransaction.js
frozeman: Technically all payload should be he. So please use a Regex `/[a-f0-9]/i.test(val)` and test if it works.

holiman:
All good with the else-block which throws, I'd also have been fine with coercing into string. Just some minor comments.

luclu:
Thank you Martin for the feedback, I made some changes for a stricter validation.
Also next week I will have a look at the checksums.

alexvandesande:
Luca, would it be possible to add a test for this attack?

luclu:
@alexvandesande in-browser-test added!

modules/ipc/methods/eth_sendTransaction.js
```diff
@@ -36,12 +36,13 @@ module.exports = class extends BaseProcessor {
...
                 _.each(payload.params[0], (val, key) => {
                     // if doesn't have hex then leave
-                    if (_.isString(val)) {
-
+                    if (!_.isString(val)) {
+                        throw this.ERRORS.INVALID_PAYLOAD;
+                    } else {
                         // make sure all data is lowercase and has 0x
-                        if (val) val = `0x${val.toLowerCase().replace('0x', '')}`;
+                        if (val) val = `0x${val.toLowerCase().replace(/^0x/igm, '')}`;
```
evertonfraga: in this context, the `g` and `m` flags seems to be redundant. As the regexp starts with `^`.

luclu: good point!

tests/mocha-in-browser/spec/permissions-spec.js
```diff
@@ -51,20 +51,21 @@ describe("Permissions", function() {
...
                     expect(window.permissions.accounts).to.include(account);
                 });
 
-                if(count === 2)
+                if (count === 2) {
                     done();
+                }
 
-                count++;
+                count += count;
```
alexvandesande: Why do you double count at each round?

Issue #1781, opened 2017-03-21
Restructure IPC endpoint logic


luclu:
### Rationale 

Currently Mist will default to `~/Library/Ethereum/geth.ipc` (path will differ for other OS).
If the user should run a geth externally with a custom ipc path like `geth --ipcpath ~/test.ipc` this Mist must be appended by the flag `Mist --rpc ~/test.ipc`. 

*Note: a GUI option to set change the node-dir and IPC endpoint will be worked on soon*

As the IPC endpoints in mainnet mode (`~/Library/Ethereum/geth.ipc`) and the testnet mode (`~/Library/Ethereum/testnet/geth.ipc`) differ Mist won't connect to an externally started geth node in testnet mode. 
### Proposal
To improve this situation Mist should try to connect to different endpoints:
1. if a custom endpoint is set (using commandline flag/GUI) **only** try this endpoint 

otherwise:
1. mainnet endpoint (default)
1. testnet endpoint

### Specification
On every connection try do the following checks:
1. does the endpoint exist? (geth will create and remove the endpoint with its lifecycle, though a crash/kill will leave the dead endpoint in existence)
2. try to connect to the RPC endpoint. 
3. if connection got refused check next endpoint **otherwise** set endpoint to valid for the current lifecycle of Mist and exit check routine

Issue #1735, opened 2017-03-14
current state of i18n


luclu:
Investigating, tracking and structuring issues regarding the i18n translation system.

- [ ] non-persistant language selection (resets to `default` after restart)

h3ll0fr13nd:
For some languages, let's take nl for example as for that I was testing a bit, numeral expects 'nl-nl' and is currently receiving only 'nl'. I've tried to use nl-nl but then i18next gives issues as it seems to be expecting nl.

luclu:
Thank you @h3ll0fr13nd, this has been fixed in https://github.com/ethereum/mist/pull/1783 and will in the next release.

h3ll0fr13nd:
Ok, great! Is that in another branch than the develop? I've just cloned and run the develop branch and still got nl passed to numeral instead of nl-nl: 
[2017-04-18 13:37:03.186] [INFO] ipcCommunicator - Backend language set to:  nl
[2017-04-18 13:37:03.236] [ERROR] (ui: mist) - numeral.js couldn't set number formating: Unknown language : nl

luclu:
Ah yes, you are right! This error appears as there is no `nl` config upstream. It will disappear after it has been added. As this is not really an error I raised https://github.com/ethereum/mist/pull/2154.

PR #2093, opened 2017-03-31
[Settings] use loki to store settings; restructure IPC endpoint logic; i18n persistence

luclu:
### loki storage

I propose to use the collection `SYS_config` in such a way, that we store the settings object as the first element of the `data` array (`coll.data[0]`, retrivable via `coll.get(1)`). `Settings.loadConfig(key)` and `Settings.saveConfig(key, value)` will get/create the according structure within this object. While using those lowlevel setters/getters might be possible, using concrete setters/getters for the key/value will allow to evaluate additional parameters like commandline flags and set defaults.
- [x] use loki (collection: `SYS_config`) instead of file based (`node`, `network`, etc.) config storage
- [x] add in `module/settings.js` (needed to change some load order in `main.js`)
  - [x] add lowlevel `Settings.loadConfig(key)` and `Settings.saveConfig(key, value)` as well as
  - [x] getter/setter for 
    - [x] `Settings.nodeType`
    - [x] `Settings.network`
    - [x] `Settings.language`
- [ ] improve config structure (add `activeClient`/`preferredClient`)
- [ ] **not needed** migration logic

### IPC endpoint logic
See https://github.com/ethereum/mist/issues/1781
- [x] store default IPC paths per network in `clientBinaries.json`
- [x] check if IPC paths for each network (mainnet, testnet) exists (nodes add/remove endpoint with its lifecycle)
- [x] only probe for 3000ms on existing IPC endpoints (i.e. check if alive; node didn't crash leaving file)

### i18n
See https://github.com/ethereum/mist/issues/1735
- [x] i18n persistence via `Settings.language`
- [x] pass application language to webview 
  - [x] via *Accepted_Lang* header (test via http://www.reliply.org/tools/requestheaders.php)
  - [ ] **code available, but commented** reload browser page after language change (should we do this, or require a full restart of Mist?)
  - [x] set `navigator.language` for browserWindow
  - [ ] **code available, but commented** set `navigator.language` for console
- [x] menu rebuild improved (directly start with correct internationalization without delay; decrease delay of inactivity after node/network switch)

### clientBinaries/clientBinaryManager
requires: https://github.com/ethereum/ethereum-client-binaries/pull/13
- [x] add `--clientbinaries` flag to Mist for custom `clientBinaries.json`
- [x] dynamically populate menu entries for nodes/networks
- [x] update `menu -> account -> backup` path
- [ ] refactor update-skipping feature to allow for multiple nodes

### edge-cases
- [ ] if a dead **and** a alive IPC endpoint exist it might **not** connect to the alive one and instead start its own node / (only checks the first existing endpoint)


frozeman:
Why storing the ipc path in the client binary json?

Otherwise looks good, tho I have no time to check the code @maiavictor and @evertonfraga please check.

luclu:
> Why storing the ipc path in the client binary json?

I think this is a better solution than hardcoding them into `ethereumNode.js`, because every client as well as testnet might have a different path for the endpoint.

modules/settings.js
```diff
@@ -231,44 +241,160 @@ class Settings {
...
+
+        if (!obj) {
+            this.initConfig();
+            return this.loadConfig(key);
+        }
+
+        return this.deepEval(obj, key);
+    }
+
+    deepEval(obj, is, value) {
```
luclu: hehe, I just figured that this has the same functionality as [lodash.set](https://lodash.com/docs/#set) and  [lodash.get](https://lodash.com/docs/#get), which is available as https://www.npmjs.com/package/lodash-get-or-set. 
I will change that in hopes of stricter and more reliable validation.

main.js
```diff
@@ -446,6 +317,136 @@ onReady = () => {
     }
 }; /* onReady() */
 
+// initialise the db
```
frozeman: Your PRs are really hard to review, as you move huge code blocks around, which make it impossible for me to see what YOU changed

frozeman: This is core logic, which when changes can break the whole app. So it needs to be clear what changed, or you really know what you are doing and build that core logic. ;)

modules/clientBinaryManager.js
```diff
@@ -13,10 +13,11 @@ const log = require('./utils/logger').create('ClientBinaryManager');
 
 
 // should be       'https://raw.githubusercontent.com/ethereum/mist/master/clientBinaries.json'
-const BINARY_URL = 'https://raw.githubusercontent.com/ethereum/mist/master/clientBinaries.json';
+// const BINARY_URL = 'https://raw.githubusercontent.com/ethereum/mist/6f2adff22569447a84639314e77379852cd273bc/clientBinaries.json';
+const BINARY_URL = 'https://raw.githubusercontent.com/ethereum/mist/a145fe24e0e98bdfd36275b4e36ca10d8781639b/clientBinaries.json';
```
frozeman: Why you hardcore a custom commit?

modules/clientBinaryManager.js
```diff
@@ -13,10 +13,11 @@ const log = require('./utils/logger').create('ClientBinaryManager');
...
 
 const ALLOWED_DOWNLOAD_URLS_REGEX =
-    /^https:\/\/(?:(?:[A-Za-z0-9](?:[A-Za-z0-9-]{0,61}[A-Za-z0-9])?\.)?ethereum\.org\/|gethstore\.blob\.core\.windows\.net\/|bintray\.com\/artifact\/download\/karalabe\/ethereum\/)(?:.+)/;  // eslint-disable-line max-len
+    /^https:\/\/(?:(?:[A-Za-z0-9](?:[A-Za-z0-9-]{0,61}[A-Za-z0-9])?\.)?ethereum\.org\/|gethstore\.blob\.core\.windows\.net\/)(?:.+)/;  // eslint-disable-line max-len
```
frozeman: Whats in bintray?

modules/clientBinaryManager.js
evertonfraga: that was for testing purposes, as I checked with him

evertonfraga:
- [ ] Regenerate `clientBinaries.json`, as with the new code an exception is thrown when using the old file.

evertonfraga:
Closing for now. The idea good and we should reinvest some time into it.

PR #2096, opened 2017-03-31
[travis] only increase timeout for mac build

luclu:
This will restore the build's process real-time console output for linux and windows builds.
Otherwise the console output will only be flushed after the script finishes. And in some cases this hinders debugging (like: https://travis-ci.org/ethereum/mist/jobs/217224093#L805).


alexvandesande:
Conflicts needs solving

luclu:
@evertonfraga do you have any idea, why this is failing on all jobs?

luclu:
Wow that was stupid of me.. :D

PR #2144, opened 2017-04-13
[clientBinaryManager] prepare for GUI node selection; prefer global node if newer

luclu:
Goes with https://github.com/ethereum/ethereum-client-binaries/pull/13

**update: integreated in https://github.com/ethereum/mist/pull/2093**

### DIFF
--- a/modules/clientBinaryManager.js
+++ b/modules/clientBinaryManager.js
@@ -13,7 +13,7 @@ const log = require('./utils/logger').create('ClientBinaryManager');
 
 
 // should be       'https://raw.githubusercontent.com/ethereum/mist/master/clientBinaries.json'
-const BINARY_URL = 'https://raw.githubusercontent.com/ethereum/mist/master/clientBinaries.json';
+const BINARY_URL = 'https://raw.githubusercontent.com/ethereum/mist/0281a97d6140a6dadd4c49b668398312711f9281/clientBinaries.json';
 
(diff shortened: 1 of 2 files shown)


evertonfraga:
I think it's risky to let Mist execute binaries it hasn't control of.

we have some options here:
1) user have to explicitly start geth, and mist ignores starting its own.
2) mist downloads and executes its own clients.
3) additional checks, like code signing (not md5 only), to allow forward
compatibility. As seen on peter's example. He had a geth version that mist
wouldn't recognize on clientbinaries.

modules/clientBinaryManager.js
```diff
@@ -13,10 +13,10 @@ const log = require('./utils/logger').create('ClientBinaryManager');
 
 
 // should be       'https://raw.githubusercontent.com/ethereum/mist/master/clientBinaries.json'
-const BINARY_URL = 'https://raw.githubusercontent.com/ethereum/mist/master/clientBinaries.json';
+const BINARY_URL = 'https://raw.githubusercontent.com/ethereum/mist/f12235657927f5e420d97bffdc5da13e9b98c069/clientBinaries.json';
```
evertonfraga: @luclu why is this json locked to 1.5.9?

luclu: it's just for testing purposes

PR #2159, opened 2017-04-19
0.8.10 into master

evertonfraga:


wikiterm:
Hi,
I would like to localize the Mist browser into Arabic. How does it work?
Regards,
WT

PR #2224, opened 2017-05-08
[travis] fix github uploader (github API behaviour changed)

luclu:
Apparently Github's API now does require to re-set the `tag_name` when patching a release drafts text.

Funnily this is not necessary for my fork branch (see https://travis-ci.org/luclu/mist/jobs/229764306) but the main repo.


evertonfraga:
For organization sake i think it should be merged to develop first

luclu:
You are right. I only raised it against master the get travis to upload directly without this even be merged. :)

PR #2227, opened 2017-05-09
Fixes wallet header

evertonfraga:
By adding back mistMode on browser preloader.

Old
image

New
image

### DIFF
--- a/modules/preloader/browser.js
+++ b/modules/preloader/browser.js
@@ -26,6 +26,7 @@ process.on('loaded', function () {
 
 
 window.mist = mist();
+window.mistMode = 'mist';
 
 // prevent overwriting the Dapps Web3
 delete global.Web3;


evertonfraga:
@frozeman was there any reason for `mistMode` not be in this scope?

frozeman:
The reason is that the wallet has its own preloader, and as we made the wallet also work for full wallet mode i temporarily added the `mistMode`

My idea was, once we drop the full wallet mode, we make the large header the default.
So i would rather change that one the wallet, then adding a non standardised `mistMode` to other preluders, which i planned on removing all together.

Better would be to put something in the `walletMain.js` so that it stays like the small header and the default when the right tag is present is the appheader. Then we can remove the mist tag.
It would break older versions of mist, tho.

We should be careful what we expose to dapps, not that they start to rely on it ;)

evertonfraga:
we could do that via user agent, but personally I don't like it. Another possible solution is to add some properties to the already existing `navigator` object, on the dapps preloader. 

`navigator.URLBarFullMode = true`

PR #2327, opened 2017-05-23
i18n korean translation

gnujoow:
here is pr for 18n korean :)
@atomrigs


gnujoow:
thanks for merge this pr :)

PR #2723, opened 2017-07-06
Adding rinkeby support

evertonfraga:
As a bonus, user can also choose a local private network.


interface/client/templates/popupWindows/onboardingScreen.html
```diff
@@ -46,7 +46,7 @@
                     </p>
 
                     {{#if TemplateVar.get "testnet"}}
-                        <p>{{{i18n "mist.popupWindows.onboarding.mineItDescription"}}}</p>
+                        <p>{{{i18n "mist.popupWindows.onboarding.FaucetDescription"}}}</p>
```
frozeman: I would have made that camelCase

evertonfraga: @frozeman done.

interface/client/templates/popupWindows/onboardingScreen.js
```diff
@@ -81,7 +90,7 @@ Template['popupWindows_onboardingScreen'].helpers({
 
     @method syncStatus
     */
-    'syncStatus': function () {
+    syncStatus() {
```
frozeman: Thats ES6 right here, if we would remove the ecmascript package, this would break.

modules/ethereumNode.js
```diff
@@ -350,14 +361,39 @@ class EthereumNode extends EventEmitter {
...
-                if (network == 'test') {
-                    args = (nodeType === 'geth')
-                        ? ['--testnet', '--fast', '--ipcpath', Settings.rpcIpcPath]
-                        : ['--morden', '--unsafe-transactions'];
-                }
-                // START MAINNET
-                else {
+                switch (network) {
+                // STARTS ROPSTEN
+                case 'test':
```
frozeman: why we call this "test" and the other by its name? To be backwards compatible?

frozeman:
I didn't test it on my machine, but it looks good from the code changes.

modules/ethereumNode.js
evertonfraga: There are some reasons:
- backwards compatibility, yes
- Geth refers to the networks as `--testnet` and `--rinkeby`, so we're actually being consistent with their flags
- forward compatibility: that'll work even if the testnet flag changes to a different chain, so we're not being specific to Ropsten in the internals, only when it's displayed.

PR #2752, opened 2017-07-12
Wallet header fix

frozeman:
Fixes the wallet header for the soon to be released new wallet.
It only sets the `window.mistMode = 'wallet'` on the wallet in standalone mode, where the wallet then only shows the small header. Otherwise it will always show the large header when in mist.

Network info is never shown in mist, expect in the standalone mode, and in external browsers without the `mist` API object.

As the wallet is centrally hosted it should be working also on old mist versions, as its not expecting anymore any special set variable, expect for the standalone mode, where its bundled anyway.


frozeman:
NOTE: I Also changed the name of the "private network", to "solo network". As this private network is not able to connect to other nodes

PR #2792, opened 2017-07-18
Add bzz and .eth to urls

alexvandesande:
If the user just types a word on the URL bar, without any protocol or TLD, then default it to bzz://xxxxx.eth

### DIFF
--- a/interface/client/lib/helpers/helperFunctions.js
+++ b/interface/client/lib/helpers/helperFunctions.js
@@ -79,22 +79,21 @@ Format Urls, e.g add a default protocol if on is missing.
 @param {String} url
 **/
 Helpers.formatUrl = function (url) {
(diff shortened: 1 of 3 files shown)


frozeman:
Please remove the `console.log`s. Also travis is still failing, any progress on that @evertonfraga ?

alexvandesande:
@frozeman these particular console.logs are not just remainders, but they were put in practice: if you try to upload a file that isn't a webpage or an image, uploading them to swarm will never show you the url, but directly show a download image. So until we add a intermediary download page to electron, that console is the only thing that will show the url

evertonfraga:
On mac those tests pass: https://travis-ci.org/ethereum/mist/jobs/255048562

Also locally. I am re-running the job on travis. Let's see what happens.

PR #2934, opened 2017-08-13
Light client

Ewgen9:


evertonfraga:
Hi @Ewgen9, thanks for the pull request. 

I wish this change would be that simple, but the thing is the light client is not yet fully compatible with all the features Mist and Wallet needs.

That said, the light client won't be a drop-in replacement to `--fast`.

evertonfraga:
Duplicated of https://github.com/ethereum/mist/pull/1547

PR #3000, opened 2017-08-29
Update copyright year 2016 to 2017 on About page

dongsam:
Update copyright year `2016` to `2017` on `About Mist` page


evertonfraga:
Thanks @dongsam. It'll be on the following version.

PR #3029, opened 2017-09-08
Fixing bug in URL formatter

evertonfraga:
It wasn't accepting `localhost` urls. with or without port.


alexvandesande:
This works for me and is a simple enough fix. LGTM.

evertonfraga:
We should check this later in the future.

PR #3034, opened 2017-09-11
Compiles ES6+

marcgarreau:
#### What does it do?
Adds the ability to write ES6+ features in electron (async/await, module imports, etc.)
- ~~The `.compilerc` file is picked up by the electron-compile and can be further tweaked. See: https://github.com/electron/electron-compile#how-do-i-set-up-babel--less--whatever-the-way-i-want~~
- Implementation updated to use a simpler configuration with Babel.
- In `menuItems.js`, `devtToolsSubMenu` was never instantiated with `const` or `let`, so I did to resolve a new error I received.
- There's a small async/await example in main.js to verify nothing explodes while testing.
Small diff. Take it from the top.
- ~~`electron-compile` & `electron-compilers` to JIT compile ES6+ within electron. See: https://github.com/electron/electron-compile~~
- `babel-gulp` for prod transpiling
- `babel-register` for dev transpiling
- `babel-preset-es2016` - es7 preset


evertonfraga:
Message from CI:

```
Unhandled rejection Error: Error: Application entry file "entry.js" in the "/home/travis/build/ethereum/mist/dist_mist/dist/linux-unpacked/resources/app.asar" does not exist. Seems like a wrong configuration.
```

marcgarreau:
@evertonfraga yup - whoops. Didn't have all the build steps accounted for (and tools installed); waiting through that now. Will update when finished 😎

evertonfraga:
No problem. Breaking CIs are a good thing  :)

I can take a look at this tomorrow morning, don't worry.

evertonfraga:
I feel way more comfortable with babel rather than electron-compiler.

nbaulesglobalsolutions:
[Text File.txt](https://github.com/ethereum/mist/files/1786595/Text.File.txt)

PR #3042, opened 2017-09-14
Updating geth to 1.7.0 for linux&win

sakamoto1255:
Just for fun though.

#### What does it do?
Updates geth to 1.7.0 LOLOLOLOLOL
see https://geth.ethereum.org/downloads/


sakamoto1255:
Geth for mac released . I will add though

PR #3063, opened 2017-09-20
Core intro

marcgarreau:
*[ in progress / up for discussion ]*

#### What does it do?
Introduces a Redux store to begin tracking global state.
- This initial effort is a proof-of-concept to capture some of the activity and state. No new functionality is added; effectively we're just logging out activity and state.
- The diffs in main.js look large, but it's not that dramatic. Everything has been nested one level to reside inside an `init` script. The purpose is to allow control flow using `async/await`.
- Once a Redux store is created, actions can be dispatched to it. Those actions flow through the reducers and ultimately produce a new global state.
- The Redux DevTools can be used to see actions as they're dispatched, and how they impact the global state. (See screenshot below for an example.)
- Right now, the global state is broken up into two namespaces: `settings` and `ui`. There's a lot shoved into both of those. This can be further broken down into smaller slices, as appropriate (e.g. namespaces for database specific items, or i18n, or whatever makes sense).
- To view Redux DevTools, visit: http://remotedev.io/local/ (or run `yarn dev:tools`). In a webapp environment, these tools could be viewed in the Chrome tools, but using the tools remotely is necessary, given the sandboxed electron environment.
- Redux isn't a huge win for the backend yet, besides being able to visualize global state. The backend just needs to execute its list of commands in a mostly synchronous fashion to kick off the app. Redux especially excels with handling user interaction, which would come later and build off this foundation.
The `modules/core/` directory has the new Redux logic. There's only one dead-simple action example for now: quitting the app. The store listens for that particular update, then shuts the app down.
`redux` - global state management
`redux-thunk` - not used yet, other than being wired up in the store as middleware, but it is the standard way to handle asynchronous activity in Redux.
`remote-redux-devtool` - broadcasts actions to the dev tools. Currently works in both prod and dev, but can be restricted to dev if desired. There's currently no harm in these actions being public.
#### Relevant screenshots?
Redux DevTools -- list of actions on left, global state on the right
screen shot 2017-09-20 at 11 59 15 am
Redux DevTools -- selected action on left, body of the action on right
screen shot 2017-09-20 at 12 00 09 pm
Redux DevTools -- selected action on left, diff of global state change as a result of the action on right
screen shot 2017-09-20 at 12 00 19 pm


marcgarreau:
closing to start work on a dedicated branch

PR #3067, opened 2017-09-22
Update app.it.i18n.json

hacktar:
#### What does it do?
Fix translation

stayAnonymous and authorize now translated


alexvandesande:
Thanks for the PR!

PR #3068, opened 2017-09-22
Updating geth

Horbi:


alexvandesande:
Thanks for that PR, @Horbi, but we have already updated Geth to 1.7.0: https://github.com/ethereum/mist/blob/master/clientBinaries.json

Any particular reason you wanted 1.6.7?

PR #3130, opened 2017-10-05
Refactors window logic/state into module

marcgarreau:
#### What does it do?
Refactors Window logic into the windows module. 
There's duplicate Window state throughout the application, where ever a new window needs to be created. The goal is to hide as much of that away as possible within the Window module.
- In `main.js`, you can see how much cleaner it is to create windows.
- The refactored logic is in `/modules/windows.js`. Now all default settings are in a method called `getDefaultOptionsForType`.

### DIFF
--- a/interface/client/templates/popupWindows/about.html
+++ b/interface/client/templates/popupWindows/about.html
@@ -10,7 +10,7 @@
                 License {{mist.license}}<br>
                 GitHub <a href="https://github.com/ethereum/mist" target="_blank">github.com/ethereum/mist</a>
             </p>
-            <small>Copyright 2016 Ethereum Foundation</small>
+            <small>Copyright 2017 Ethereum Foundation</small>
         </div>
     </div>
 </template>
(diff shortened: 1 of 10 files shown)


frozeman:
As you are already on this, it would make sense to improve the popup window logic by pre-creating 5 windows, and only reloading the content, to be able to show them faster and create a more native feeling user experience.

marcgarreau:
@frozeman good hypotheses - I'll start digging around.

interface/client/templates/popupWindows/about.html
```diff
@@ -10,7 +10,7 @@
                 License {{mist.license}}<br>
                 GitHub <a href="https://github.com/ethereum/mist" target="_blank">github.com/ethereum/mist</a>
             </p>
-            <small>Copyright 2016 Ethereum Foundation</small>
+            <small>Copyright 2017 Ethereum Foundation</small>
```
shirokoff: It should be on store

marcgarreau:
closing PR while rethinking strategy; will reopen this or a new PR when the dust settles.

PR #3186, opened 2017-10-20
Generic window refactor

marcgarreau:
#### What does it do?
Refactors the use of electron browser windows for better performance.
- Meteor startup times appear to be the biggest detractor from performance within the app, specifically when opening any new window. This refactor introduces a reusable ("generic") window for popup windows to use if it's available.
- At the moment, the generic window has web3 capabilities by default, given the use of preloaders and the uncertainty of what the window will be used for. If this is too great a security risk, a future iteration can introduce a second generic window for use with popups that don't require web3.
- There's only one generic window for now. If multiple popups are opened, the existing implementation takes over (loading window, then new Meteor app).
`windows.js` is the bulk of it. There's a comment trail along the way.
#### Screenshots
Redux tracked state:
screen shot 2017-11-14 at 9 23 49 am

### DIFF
--- a/interface/client/appStart.js
+++ b/interface/client/appStart.js
@@ -74,7 +74,7 @@ Meteor.startup(function () {
         mistInit();
     }
 
-    if (store) { store.dispatch(getLanguage()); }
+    store.dispatch(getLanguage());
 
     // change moment and numeral language, when language changes
     Tracker.autorun(function () {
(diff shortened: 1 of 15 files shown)


evertonfraga:
I get an error in Check for Updates window, and it does not load properly:

image

PR #3200, opened 2017-10-23
Fix encoding error(KOR)

wjcloud:
#### What does it do?
Fix encoding error - this will enable korean for ethereum installer.


evertonfraga:
That is great, @wjcloud.

I'll try to fix other languages as well.

Issue #2713, opened 2017-07-05
mist.lokidb has a growing number of duplicated information


evertonfraga:
The key `collections.DynamicViews` has the same dictionary appended 387 times, in my environment.

```
const sortedTabs = Tabs.addDynamicView('sorted_tabs');
```
https://github.com/ethereum/mist/blob/develop/main.js#L510

 
```
    {  
               "collection":null,
               "name":"sorted_tabs",
               "rebuildPending":false,
               "options":{  
                  "persistent":false,
                  "sortPriority":"passive",
                  "minRebuildInterval":1
               },
               "resultset":{  
                  "collection":null,
                  "filteredrows":[  
                     0,
                     1
                  ],
                  "filterInitialized":true
               },
               "resultdata":[  

               ],
               "resultsdirty":true,
               "cachedresultset":null,
               "filterPipeline":[  

               ],
               "sortFunction":null,
               "sortCriteria":[  
                  [  
                     "position",
                     false
                  ]
               ],
               "sortDirty":false,
               "events":{  
                  "rebuild":[  

                  ]
               }
            }
```

evertonfraga:
Fixed in #3288.

PR #3288, opened 2017-11-15
Prevents duplicate Dynamic View entries

marcgarreau:
#### What does it do?
Addresses #2713 - prevents duplicate Dynamic View instances from stacking up in the local db.


evertonfraga:
Great, Marc

PR #3291, opened 2017-11-15
Auto-generating SHA256 checksums

evertonfraga:
#### What does it do?
It runs sha256 against Mist/EW artifacts and appends them to the release notes if there are any release drafts.

We used to distribute binaries with MD5 checksums. SHA-256 checksums were generated manually until now. 

-

No.

#### Relevant screenshots?
This is what we aim for. We can't ensure what order the builds will finish, so it still requires a little text manipulation at the end, as all builds will print out the table header as well.

image

### DIFF
--- a/gulpTasks/publishing.js
+++ b/gulpTasks/publishing.js
@@ -21,16 +21,11 @@ gulp.task('checksums', (cb) => {
     let command;
     let argument = '';
 
-    switch (process.platform) {
-    case 'darwin':
-        command = 'md5';
-        break;
-    case 'win32':
+    if (process.platform === 'win32') {
         command = 'certUtil -hashfile';
-        argument = 'md5';
-        break;
-    default:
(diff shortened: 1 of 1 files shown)


Taledeyo:
I try to get wallet.

gulpTasks/publishing.js
```diff
@@ -90,14 +85,22 @@ gulp.task('upload-binaries', (cb) => {
...
+                                const line = e.replace('\n', '').split('  ');
+                                return `<sub>${line[1]}</sub> | <sub>\`${line[0]}\`</sub>`;
+                            }).join('\n'))
                         })
                     });
                 }
+            })
+            .catch((err) => {
+                console.log(err);
             });
```
ja1977ja: eba69844bc1d8e2ba5253f016c7e65815ce66e892b85fb3d6533babcd38c7376

Issue #3384, opened 2017-12-06
Geth Version Update Screen Not Allowing for Copy and Paste of Checksum (Windows)


gabefair:
- [ ] I've asked for help in the [Mist Gitter](http://gitter.im/ethereum/mist) before filing this issue.

```
Version: `0.9.3`
OS & Version: windows
Node version: `geth 1.7.0` 
Number of blocks synchronized: 0
```

I want to copy and paste the checksum but it isn't allowed by the GUI.

![image](https://user-images.githubusercontent.com/2096785/33685344-44b4a3b4-da9f-11e7-8c06-43468559e74f.png)

PR #3396, opened 2017-12-08
Make individual text selectable in popup

tobiv:
#### What does it do?

Make individual text selectable in the popup-window, for example to copy the checksum of an update.

See issue https://github.com/ethereum/mist/issues/3384


evertonfraga:
Thanks for the contribution! :)

PR #3418, opened 2017-12-12
Upgrade del to 3.0.0

ryanio:
I was running into an error with del when running the clean-dist task in gulp: `Error: EINVAL: invalid argument, rmdir`

Upgrading to del 3.0.0 fixed for me.


ryanio:
Moved to #3421

Issue #3424, opened 2017-12-12
Remove Swarm from init process


evertonfraga:
As Swarm is not live yet, we can pull it from the regular initialization process. We should keep it at hand putting a toggle option at the end of the Develop menu.

### Develop menu
(…) all options we have today
`---------------` (separator)
[ √ ] Enable Swarm

- [x] Remove swarm from the init process.
- [x] Add toggle item on Develop menu.
- [x] Enable `File > Upload to swarm` only if Swarm is on.
- [x] Think of the best way to inform the user if it tries to access a `bzz://` url.
- [x] Persist choice using lokiDB

ryanio:
Sounds good.

I think we'll also want to present the user connection info or errors while enabling. Not sure if we have a good spot for this, or if it warrants a simple pop-up window with the info that then hides upon success. I can play around with something simple after accomplishing the tasks above.

evertonfraga:
Sure thing. 

Thinking of a general solution, we could present a window containing live logs. Aiming for simplicity, we could even fire up a terminal with `tail -f [logfile]` (and equivalents for all OSes). This could be a quick way to display what's happening to the users.

AkliOnet:
A nice terminal ...

PR #3431, opened 2017-12-14
Remove swarm from init

ryanio:
Continuing discussion from #3424:

Everything is working here on my comp:

- [x] Remove swarm from the init process.
- [x] Add toggle item on Develop menu.
- [x] Enable `File > Upload to swarm` only if Swarm is on.
- [x] Think of the best way to inform the user if it tries to access a `bzz://` url.
- [x] Persist choice using lokiDB

I also added:

- [x] Run Mist with --swarm to run on start up regardless of lokiDB setting. --swarm=false to ensure it doesn't run.

We're logging relevant swarm data, but I would love to surface some more info to the end user for interaction/debugging purposes. Will keep thinking on it:

- [x] Show swarm status/info to the user (on enable, disable, upload, etc.)
  - (Solved for now by making access to logs easier in #3434)


evertonfraga:
great PR. I'll review after Marc's.

I can show you tomorrow what I've been thinking of to inform users of these questions you've raised.

modules/swarmNode.js
```diff
@@ -86,14 +86,23 @@ class SwarmNode extends EventEmitter {
         return this.startUsingGateway();
     }
 
+    stop() {
+        if (!this._swarm) {
+            return Q.reject(new Error('Swarm not initialized. Have you called swarmNode.init()?'));
```
marcgarreau: nit: the error verbiage looks like logging for your debugging purposes. I'd leave out the implementation detail (`swarmNode.init()`) for production.

main.js
```diff
@@ -161,6 +160,14 @@ onReady = () => {
 
     // Enable the Swarm protocol
     protocol.registerHttpProtocol('bzz', (request, callback) => {
+        if (!store.getState().settings.swarmEnabled) {
+            let error = global.i18n.t('mist.errors.swarm.notEnabled');
+            dialog.showErrorBox('Error', error);
+            callback({ error: error });
+            store.dispatch({ type: '[MAIN]:PROTOCOL:ERROR', payload: { protocol: 'bzz', error: error } });
```
marcgarreau: nit: with ES6, you can reduce identical object key/value pairs to just the key.
`callback({ error });`
`... payload: { protocol: 'bzz', error } ...`

ryanio: nice, that's awesome. thanks for the tip

marcgarreau:
1) starting and stopping swarm appears to work well locally.
2) I don't know how important this feature actually is, but the enabled-on-start setting isn't picked up if you restart after turning swarm on. I assume this is still the result of not populating the Redux initial state with the value (and is probably already on your radar).

ryanio:
1. Great
2. Hm well if we want it to start for tests, in main.js it checks `Setting.enableSwarmOnStart`. So we could make sure that value returns true in the test runner.

marcgarreau:
@ryanio sorry, that was unclear. Swarm should remain off for the integration tests. Just pointing out that whenever reopening Mist, Swarm is off by default, even if you last closed Mist with it enabled.

ryanio:
@marcgarreau Ah I see what you mean now. I think you're right, I just pushed a [commit](https://github.com/ethereum/mist/pull/3431/commits/35ca6db8e370865441e795148989be2dd90c1b0e) that should fix it by syncing the Redux state on getting the setting (so ugly😩, can't wait to be fully on redux). Can you let me know if it works for you now?

modules/core/settings/actions.js
```diff
@@ -88,3 +96,74 @@ export function getLanguage(event) {
         }
     }
 }
+
+export function toggleSwarm(event) {
+    return (dispatch, getState) => {
+        if ([SwarmState.Enabled, SwarmState.Enabling].indexOf(getState().settings.swarmState) > -1) {
```
marcgarreau: ES7(?) added the `includes` method on the Array prototype. so,
`[x, y].includes(getState().settings.swarmState)`
cleans that up a little.

modules/core/settings/actions.js
```diff
@@ -88,3 +96,74 @@ export function getLanguage(event) {
...
+            Settings.enableSwarmOnStart = true;
+            dispatch({ type: '[MAIN]:SWARM:ENABLE_ON_START' });
+        }
+    }
+}
+
+export function setSwarmEnableOnStart() {
+    return (dispatch, getState) => {
+        dispatch({ type: '[MAIN]:SWARM:ENABLE_ON_START' });
+    }
```
marcgarreau: Whenever you don't need to perform any side-effects, you can just use a synchronous action creator. i.e.
```
export function setSwarmEnableOnStart() {
   return { type: '[MAIN]:SWARM:ENABLE_ON_START' };
}
```

modules/menuItems.js
```diff
@@ -547,6 +549,20 @@ let menuTempl = function (webviews) {
...
+        type: 'separator'
+    },
+    {
+        label: i18n.t('mist.applicationMenu.develop.enableSwarm'),
+        enabled: true,
+        checked: [SwarmState.Enabling, SwarmState.Enabled].indexOf(global.store.getState().settings.swarmState) > -1,
+        type: 'checkbox',
+        click() {
+            store.dispatch(toggleSwarm());
+            store.dispatch(toggleSwarmOnStart());
```
marcgarreau: I like the idea of having `toggleSwarm` take care of the `onStart` stuff. From a dev's perspective I think It would be nice not to have to care/know about this implementation detail at this level. What do you think?

main.js
```diff
@@ -302,11 +294,9 @@ onReady = () => {
             return ethereumNode.init();
         })
         .then(() => {
-            // Wallet shouldn't start Swarm || TODO: TravisCI failing to download Swarm
-            if (global.mode === 'wallet' || Settings.inAutoTestMode) {
```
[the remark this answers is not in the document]
ryanio: Sure I can do that, I'll just add it into the getter in `settings.js` like so:
```js
    get enableSwarmOnStart() {
+        if (global.mode === 'wallet') {
+            return false;
+       }
+
```

Should the wallet also not have `Upload to Swarm` and `Enable Swarm` in the menu bar then?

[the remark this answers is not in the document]
evertonfraga: yep, it shouldn't present those options there.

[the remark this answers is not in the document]
ryanio: great, 633875b should take care of it :)

evertonfraga:
Oh, I also need to reproduce this behavior here. It fired an error where it shouldn't:
screen shot 2017-12-16 at 11 34 25 am

evertonfraga:
This PR is looking great so far @ryanio. Other than that aforementioned behavior, it's good to merge.

modules/menuItems.js
```diff
@@ -224,40 +264,9 @@ let menuTempl = function (webviews) {
...
-                              }});
-                              LocalStore.set('selectedTab', 'browser');
-                            `);
-                            swarmLog.info('Hash uploaded:', hash);
-                        }).catch(e => swarmLog.error(e));
-                    }
-                }
-            }]
-    });
+            ...swarmUpload
```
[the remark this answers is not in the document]
ryanio: Just to unpack the array of the [seperator, swarmUpload_menu_item] into the submenu where it’s located. And if it’s null (not set because in wallet mode) then it doesn’t add any items.

ryanio:
Thx!

That error may happen if you have a tab in mist already open with a bzz:// protocol and then you open mist without swarm enabled. Is this your case by any chance?

However now thinking about it, if you have a bzz tab open, since we’re not async waiting for swarm to start in splash now until loading the tabs, this could happen even with swarm enabled. Let me think about a possible solution.

evertonfraga:
It looks good from what I've read. I'll fiddle with this tomorrow.

PR #3550, opened 2018-01-11
Update app.ko.i18n.json

GoopyAspirin:
#### What does it do?
Replaced by the correct Korean word for the situation.


evertonfraga:
Thank you spongebob!

PR #3691, opened 2018-02-16
Removes popup code

alexvandesande:
This removes the pop up code, which previously would allow any link with attribute target=_popup to open an external site within an electron window. This was only used for the shapeshift integration, but it could be a potential vector.

All shapeshift pop ups have been removed, and replaced by opening a new browser (default browser) page with changelly link.

Also, updates identicon package.


evertonfraga:
It seems there are several changes in one PR. As we're removing the onboarding screen, do you prefer making those changes separately?

alexvandesande:
I can wait until the on boarding is removed and then I'll rebase

alexvandesande:
Depends on https://github.com/ethereum/mist/pull/3697

PR #3697, opened 2018-02-19
Removes onboarding

marcgarreau:
#### What does it do?
Removes the onboarding experience.
- A vastly improved geth syncing experience mostly eliminates the need to 'entertain' the user while they wait for syncing.
- This onboarding has some quirks, like asking you which network you want to join, after you've already begun syncing one of them.
- Remote node integration (coming soon) will remove the splash screen experience altogether.
- We set the default tab to be the wallet (instead of browser - ethereum.org), so you can see first thing that you don't have an account and need to create one. The wallet UI does a clear job of spelling this out.
- In the hopefully near future, we will design more modern user onboarding/education material throughout Mist.


interface/client/appStart.js
```diff
@@ -56,12 +56,7 @@ mistInit = function () {
                 }
             });
 
-        // Sets browser as default tab if:
-        // 1) there's no record of selected tab
-        // 2) data is corrupted (no saved tab matches localstore)
-        if (!LocalStore.get('selectedTab') || !Tabs.findOne(LocalStore.get('selectedTab'))) {
-            LocalStore.set('selectedTab', 'wallet');
-        }
+        LocalStore.set('selectedTab', 'wallet');
```
evertonfraga: With this change, if the user has another tab selected and quits Mist, it'll always select the wallet.

It should keep the previous tab.

marcgarreau: i guess the logic was correct after all; reverted.

alexvandesande:
Shouldn't you also be removing `mist/interface/client/templates/popupWindows/onboardingScreen.html` and `mist/interface/client/templates/popupWindows/onboardingScreen.js`?

marcgarreau:
did some meaningful refactoring after your approval, please test before merging @evertonfraga 🙏

alexvandesande:
screen shot 2018-02-23 at 4 50 26 pm

Testing it. Seems to be working fine, but this is the screen the person sees when they click "add account". Should we add a logic to that screen where if you have no accounts it goes directly to the "create account" screen?

alexvandesande:
also, another weird behavior: even **after** I added an account it still not visible on that window until I refreshed the whole Mist UI. We need to investigate that to understand why it's not refreshing automatically.

screen shot 2018-02-23 at 4 53 03 pm

marcgarreau:
@alexvandesande 👍 will poke at both

alexvandesande:
I'm still getting the issue where it needs to refresh the whole browser in order for it to see the new accounts. But maybe we could merge this PR, which works fine, and try to solve that in a separate thread? Or just force a browser refresh after adding a key?

PR #3723, opened 2018-03-02
Adding unit tests to Travis job

evertonfraga:
#### What does it do?

It runs `yarn test:unit:once` in Travis jobs.

It should only run on Linux builds and its return codes must be correct (non-zero for errors and zero for success)


marcgarreau:
_patiently waits for the failure._ Travis has partially degraded service atm.

PR #3729, opened 2018-03-05
Updates Geth to "Frost"

evertonfraga:
#### What does it do?

Updates geth in ethereum node manifest `clientBinaries.json` to Frost version (1.8.2).

1. Run a local server to serve `clientBinaries.json`.

```
$ cd mist
$ http-server
```

2. Replace `modules/clientBinaryManager.js:16` to the manifest file path. `http://localhost:8080/clientBinaries.json` (or proper port given by previous command).

#### Relevant screenshots?

screen shot 2018-03-05 at 3 06 00 pm

### DIFF
--- a/clientBinaries.json
+++ b/clientBinaries.json
@@ -1,15 +1,15 @@
 {
     "clients": {
         "Geth": {
-            "version": "1.8.1",
+            "version": "1.8.2",
             "platforms": {
                 "linux": {
                     "x64": {
                         "download": {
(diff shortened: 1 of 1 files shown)


evertonfraga:
The persistence of the number of downloaded state entries is working like a charm. That alone would already help provide a less confusing experience to the users. 

Now, if I interrupt and start syncing again, it'll show the amount of "chain structure" items downloaded.

screen shot 2018-03-05 at 7 01 47 pm

PR #3755, opened 2018-03-15
Prettier files

ryanio:
#### What does it do?
Formats `js`, `less`, and `md` files using [prettier](https://github.com/prettier/prettier) and adds a precommit hook to automatically process files.

For development, it is recommended to add prettier to [your favorite editor to run on save](http://prettier.io/docs/en/editors.html). It makes writing code much more enjoyable.

See: [Why Prettier?](https://prettier.io/docs/en/why-prettier.html)

Reviewers should bootup mist as normal and ensure everything is working as expected.

`prettier pretty-quick husky` for running files through prettier on precommit

#### Notes
* Can add paths to ignore in .prettierignore
* One-line convenience cmd: `prettier --write "**/*.{js,less,md}"`


marcgarreau:
:hugefan: of this 🎉 

travis is being grumpy about that unzip package: `Error: Cannot find module 'unzip'`

marcgarreau:
This is not universally true, but prettier often outright replaces eslint. May be worth checking out here, otherwise in a future PR.

PR #3766, opened 2018-03-20
Refactoring upload-queue task; Prettier code style changes

evertonfraga:


ryanio:
Hm not sure why your prettier defaulted to double quotes, it should have picked up the `.prettierrc`

You can also use prettier with `--single-quote`

evertonfraga:
Yeah, I have installed prettier-atom in my editor but it didn't read from the `.prettierrc` file. Anyway, I've force-enabled the use of single quotes. Let me know if my editor is breaking another set of rules.

PR #3773, opened 2018-03-22
Remove web3.admin

ryanio:
Talked with @frozeman and it doesn't look like web3.admin is being used anywhere so this PR is to remove it.

The reason this came up is because it's throwing errors with web3 1.0 since the formatters in web3._extend were moved/removed.


marcgarreau:
`modules/preloader/popupWindows.js` still has a reference to web3Admin. Breaks for Mist.

alexvandesande:
I found only a passing reference to it on the wallet: https://github.com/ethereum/meteor-dapp-wallet/blob/develop/app/client/lib/appStart.js#L93

Seems it's just an error detection code and it can be removed easily

Hasbullah110782:
Jadilah yg terdepan

PR #3775, opened 2018-03-22
Use geth personal.signTransaction

ryanio:
This code changes the sendTransaction window in Mist to use geth to first sign the transaction before broadcasting the signed tx to the network.

This allows us to broadcast the transaction however we wish in the future, like an override to send it through a remote node (infura).

To test, try sending a transaction in mist.


interface/client/templates/popupWindows/sendTransactionConfirmation.js
```diff
@@ -398,58 +398,94 @@ Template['popupWindows_sendTransactionConfirmation'].events({
...
+      pw = null;
+      TemplateVar.set('unlocking', false);
+      if (error) {
         Tracker.afterFlush(function() {
           template.find('input[type="password"]').value = '';
           template.$('input[type="password"]').focus();
         });
-        if (e.message.indexOf('Unable to connect to socket: timeout') !== -1) {
+        switch(error) {
+          case
```
marcgarreau: syntax 💣 here

ryanio:
thanks good catch, on hold while we update to beta.33 which has the method

marcgarreau:
Nearly synced and getting spammed with: `TypeError: Cannot read property 'peerCount' of undefined`

ryanio:
Merging into #3776

PR #3780, opened 2018-03-23
Adds retryConnection listener

### LINKED ISSUE #3779: retryConnection does absolutely nothing
I'm pretty sure that hitting the retry connection button from the splash screen does nothing at all.
The problem is that there is no receiver for that retryConnection "signal".

There is only the send ('retryConnection', ...) left in code, see:
https://github.com/ethereum/mist/blob/f94d8f19166a81ee71d51788052b16d845374dd6/interface/client/templates/popupWindows/splashScreen.js#L298

the code does not have any listener like this anymore:
```
ipcMain.on('retryConnection', () => {
  kickStart();
});
```

the on('retryConnection', ...) is now missing completely within the code

This problem was introduced already by this commit: https://github.com/ethereum/mist/pull/3370/

marcgarreau:
#### What does it do?
Resolves #3779; retryConnection listener didnt survive a merge or reshuffle.

### DIFF
--- a/main.js
+++ b/main.js
@@ -327,6 +327,10 @@ function checkForLegacyChain() {
 }
 
 function initializeKickStartListeners() {
+  ipcMain.on('retryConnection', () => {
+    kickStart();
+  });
+
   ClientBinaryManager.on('status', (status, data) => {
     Windows.broadcast('uiAction_clientBinaryStatus', status, data);
   });


evertonfraga:
LGTM

evertonfraga:
I am now testing thoroughly. After hitting the retry button, it didn't actually work.
https://gist.github.com/evertonfraga/31b547b9871ce52016d2acfe7db1831e

got stale on "EthereumNode - Stopping existing node: geth main"

marcgarreau:
Proving to be not a straight-forward task. Couple notes:

1) using `--node-syncmode=potato` to simulate the failure. Starting a node always calls the `stop` function first, and this function hangs indefinitely with this test case. Unsure how to test with a more 'authentic' error (or if that's even important).
2) the whole splashScreen will be removed with `layered-nodes`, rendering this effort useless.

Pausing for the week to work on higher priority items and evaluate the pace of `layered-nodes`.

ryanio:
Closing in favor of removing the splash screen in #3825

PR #3781, opened 2018-03-23
menu: there is no "Accounts" menu anymore

philsmd:
#### What does it do?

The warning about making backups of keystore files was not updated with this change of the "Accounts" menu to a "File" menu, see: https://github.com/zhangkimii/https-github.com-ethereum-mist/commit/65a9fb3213bbf820aaad85cc5bb9a56e9155c855#diff-7412f9723fd3d73999d755e7f3214656R135 , therefore this commit makes sure that the backupHint does not ask the user to click on the "Accounts" menu (which does not exist anymore) but it now correctly mentiones the "File" menu.

This commit seems to have introduced the problem:
https://github.com/zhangkimii/https-github.com-ethereum-mist/commit/65a9fb3213bbf820aaad85cc5bb9a56e9155c855

the commit message was "move swarm menu into accounts, rename it into file menu" 

I've tried to make sure that all present backupHint translations are now also reflecting the correct File-menu-label, but unfortunately there are many, many, many translations missing :(
(off-topic: I think you should somehow involve the community to help with translations by making it easy for the contributers/translaters to see which strings are not yet translated or incorrectly translated or just copies from the english text which seems very often to be the case :( . Several translation strings seem to be pretty bad or old.)

### DIFF
--- a/interface/i18n/mist.ca.i18n.json
+++ b/interface/i18n/mist.ca.i18n.json
@@ -190,7 +190,7 @@
                 "enterPassword": "Introdueix una contrasenya",
                 "repeatPassword": "Repeteix la contrasenya",
                 "creating": "Generant el compte...",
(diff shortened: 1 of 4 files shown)


evertonfraga:
The test has passed but the ci-reported didn't get the result properly. I'm gonna close and reopen the PR.

PR #3896, opened 2018-05-17
chore(package): update electron to version 2.0.1

fredbuluku:
Closes #3851


fredbuluku:
Hello Marcgarrea,

We had a package on the Ethereum/mist which had an issue for dialogue,I had requested for the merger but it did materialize,so I was asking what happened to the package after solving the conflict?

marcgarreau:
It is not as simple as updating the version number - there will be breaking changes to be fixed throughout the app.

PR #3916, opened 2018-05-25
Enable attaching geth with a private network

ryanio:
#### What does it do?
Allows you to run `geth --dev` and run Mist with `yarn dev:electron --rpc path/to/geth.ipc` and have everything work properly (detect correct network, etc.)


evertonfraga:
Looks pretty good to me! will test it thoroughly :)

PR #3922, opened 2018-05-29
Introduces local debug server

marcgarreau:
#### What does it do?
Run `yarn dev:tools` to start a local DevTools server.

`remotedev.io/local` is regularly down and can't be relied on.

[remotedev-server](https://github.com/zalmoxisus/remotedev-server) as a dev dependency


evertonfraga:
oh, thanks for that. I've experienced myself the other day.

yarn.lock
```diff
@@ -2575,8 +2637,8 @@ ethereum-client-binaries@ethereum/ethereum-client-binaries:
     lodash.isempty "^4.4.0"
     lodash.values "^4.3.0"
     mkdirp "^0.5.1"
-    node-unzip-2 "github:glebdmitriew/node-unzip-2"
     tmp "0.0.29"
+    unzip "github:glebdmitriew/node-unzip-2"
```
evertonfraga: Got this error from travis.
```
[22:24:19] Using gulpfile ~/build/ethereum/mist/gulpfile.js
[22:24:19] Starting 'test'...
module.js:549
    throw err;
    ^
Error: Cannot find module 'node-unzip-2'
```

ryanio:
Yeah I’ve been getting it as well, probably worth adding ‘node-unzip-2’ to dependencies (devDependencies?)

yarn.lock
ryanio: hm it is in `devDependencies`. i am getting `Unresolved node modules: node-unzip-2` sometimes though with electron-builder so i'll try moving it to `dependencies` and see if that fixes

PR #3929, opened 2018-05-30
Clean up and note a to do that the regular geth  proc `stderr` is come in soon

leekwiyoung:
…g in as `stderr` for some reason


marcgarreau:
If you have a bug to report, please open an issue with a description and steps to reproduce it in English if you are able. Thanks!

PR #3941, opened 2018-06-05
Fresh lock file

marcgarreau:
#### What does it do?
Blows away last lock files and reinstalls.


evertonfraga:
Why having both `package-lock.json` and `yarn.lock`?

marcgarreau:
Updated both since both were in the repo, but agreed there should only be one. Will delete `package-lock.json`, given our documentation references yarn.

PR #3966, opened 2018-06-18
Fixes routing certain jsonrpc requests to local node in production

ryanio:


modules/ipc/methods/base.js
```diff
@@ -115,18 +115,20 @@ module.exports = class BaseProcessor {
...
+    // net_peerCount | eth_syncing | eth_subscribe[syncing]
+    // and is originating from the mist interface
+    // dev: localhost:3000, production: app.asar/interface/index.html
     if (
       conn &&
       conn.owner &&
       conn.owner.history &&
-      conn.owner.history[0].startsWith('http://localhost:3000')
+      (conn.owner.history[0].startsWith('http://localhost:3000') ||
+        conn.owner.history[0].indexOf('app.asar/interface/index.html') > -1)
```
marcgarreau: String#includes is a bit cleaner, nonblocking 👍

ryanio: yeah, I tried but node wasn't having it - it was trying to treat it as Array#includes so I assumed it didn't support with String

PR #3967, opened 2018-06-19
Fixes disabled menu buttons

marcgarreau:
#### What does it do?
Uses Redux state to determine client is geth. 
ethereumNode's `_type` isn't updated before the menu button state is determined, so the buttons appear disabled. Ideally all state is derived from Redux anyway.

### DIFF
--- a/modules/menuItems.js
+++ b/modules/menuItems.js
@@ -599,10 +599,7 @@ let menuTempl = function(webviews) {
     submenu: [
       {
         label: i18n.t('mist.applicationMenu.develop.syncModeLight'),
-        enabled:
-          ethereumNode.isOwnNode &&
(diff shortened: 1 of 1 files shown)


modules/menuItems.js
```diff
@@ -601,7 +601,7 @@ let menuTempl = function(webviews) {
         label: i18n.t('mist.applicationMenu.develop.syncModeLight'),
         enabled:
           ethereumNode.isOwnNode &&
-          ethereumNode.isGeth &&
+          store.getState().nodes.local.client === 'geth' &&
```
evertonfraga: I believe the ` === geth ` is no longer necessary, as this line would only be parsed after `ethereumNode.isOwnNode` returns true. And we do not have another own client anymore.

PR #4023, opened 2018-07-24
Updates to geth 1.8.13

evertonfraga:


evertonfraga:
Please do not consider Appveyor check result, as it's not implemented for all branches yet.

Issue #4017, opened 2018-07-20
Ethereum Wallet contract IDE verification always uses the newest compiler !


toby1991:
- [ ] I've asked for help in the [Mist Gitter](http://gitter.im/ethereum/mist) before filing this issue.

```
Version: `0.11.0`
OS & Version: osx
Number of blocks synchronized: 5998269
```

# Ethereum Wallet contract IDE verification doesn't use the compiler which has been already defined in the solidity !

the "contract IDE verification" is the solidity code compiler verification.
For example, 
I was using **the contract name as the function name for the contract construct function at `pragma solidity ^0.4.21;`** and the Ethereum Wallet's inner IDE tells that **Could not compile source code. **

error message:
>  Could not compile source code. 

>  Defining constructors as functions with the same name as the contract is deprecated. Use "constructor(...) { ... }" instead.
    function MYCONTRACTNAME() {
    ^ (Relevant source part starts here and spans across multiple lines).
, 

**but this feature is not a `pragma solidity ^0.4.21;` feature.**
[https://solidity.readthedocs.io/en/v0.4.21/contracts.html?highlight=construct](https://solidity.readthedocs.io/en/v0.4.21/contracts.html?highlight=construct)
> When a contract is created, its constructor **(*a function with the same name as the contract*)** is executed once. A constructor is optional. Only one constructor is allowed, and this means overloading is not supported. 

**however, this feature is showed in the newest version of solidity compiler `pragma solidity ^0.4.24`**
[https://solidity.readthedocs.io/en/v0.4.24/contracts.html?highlight=construct](https://solidity.readthedocs.io/en/v0.4.24/contracts.html?highlight=construct)
> When a contract is created, its constructor (a function declared with the constructor keyword) is executed once. A constructor is optional. Only one constructor is allowed, and this means overloading is not supported.

So after I change the code  
` function MYCONTRACTNAME() `
to
` constructor() `
the error message disappeared.
BUT I still leave the code `pragma solidity ^0.4.21;` declared !

**I think this is a BUG**

Ali8668:
Gooood

evertonfraga:
Hello @toby1991, can you elaborate your claim? What exactly is the contract IDE verification?

toby1991:
Hi @evertonfraga ,the "contract IDE verification" is the solidity code compiler verification.
For example, 
[...]

PR #4052, opened 2018-08-06
Locking Solidity compiler version

evertonfraga:
Fixes https://github.com/ethereum/mist/issues/4017.


marcgarreau:
Not yet responsive to pragma changes:
screen shot 2018-08-14 at 9 44 54 am

evertonfraga:
@marcgarreau should be fine for another test. to avoid any warnings about constructor, run the wallet from this PR: 
https://github.com/ethereum/meteor-dapp-wallet/pull/488 (better to start it with `meteor --port=3050` and use Browse tab to access it from mist.

fredbuluku:
Very great thanks for the good job.

FredBuluku

PR #4070, opened 2018-08-13
New SendTx and TxHistory windows

ryanio:
#### What does it do?
This PR refactors `sendTransactionConfirmationWindow` into a react component called `SendTx`.

It also introduces a new Transaction History window (component called `TxHistory`).

The `txs` array in the redux store is now persisted to disk (new deps: `redux-persist`, `electron-store`, and `redux-persist-electron-storage`).

#### How to test it

There are four primary tx types handled: send ether, send tokens, execute a contract function, and deploy a new contract. Run through these scenarios. Reference the [QA Checklist](https://github.com/ethereum/mist/wiki/QA-Checklist) for a more thorough guide.

Note that you can also dispatch any custom action in the Redux DevTools to simulate error states.

#### Screenshots

_Screenshot example sending an ERC20 token:_

screen_shot_2018-08-09_at_11 58 41_am

_Gas loading/warning/error UI:_


fredbuluku:
If there’s any problems  with merging then you should not effect the  merge,but if there’s no issues with merging then you can go ahead and make changes.

marcgarreau:
re: LOC, agreed @PhilippLgh. a good chunk of that is the build artifacts for appveyor until meteor is removed, though.

interface/components/SendTx/ContextDescription.js
```diff
@@ -0,0 +1,139 @@
...
+    );
+  }
+
+  renderTokenTransferDescription() {
+    if (this.props.params.length === 0) return;
+
+    const tokenCount = this.props.params[1].value.slice(
+      0,
+      -Math.abs(this.props.token.decimals)
+    );
```
evertonfraga: The statement above does not work for fractions of tokens. Consider these test cases:
```js
5e18.toString().slice(0, -18)
// expected: '5'
// output: '5'

0.5e18.toString().slice(0, -18)
// expected: '0.5'
// output: ''
```

evertonfraga:
Context: 
- contract creation or function call 
- `value` is set to any amount > 0 (ether being spent in that tx)

In that scenario, a dapp could generate a function call transaction with an arbitrary ether amount and the user wouldn't know he/she is spending that value.

interface/components/SendTx/ContextDescription.js
marcgarreau: Good catch. Added a formatter/util function with some unit tests.

screen shot 2018-09-30 at 11 04 28 pm

evertonfraga: yay, thanks for making the formatter and a  test suite for it.

PR #4075, opened 2018-08-18
Luclu dynamic node settings

allnewso29:


evertonfraga:
@allnewso29 thanks for the pull request, but this code is pretty old and it's outdated.

PR #4135, opened 2018-11-01
Using Ledger Nano with Wan support

thesonofja:
Hello I currently have the Desktop WAN wallet with my coins in there but now want to integrate it with my Ledger Nano S. Is it true that this would be more secure than the desktop version?

 Im also confused because when I open the desktop version it always shows everything and looks like ANYONE can see and even send away my coins. It does not ask for a password. I do not remember writing down any password when I set up the wallet. All I have in my secure files is my wancoin keystore file.

Any advice and detailed instructions are appreciated!

Regards,

J


marcgarreau:
hey 👋 it seems that you're looking for a different repo - I guess a Wanchain fork of Mist. I'm going to close this PR, but please correct me if I've misunderstood.

PR #4152, opened 2018-11-19
Txhistory fixes (#4150)

evertonfraga:
* Fix undefined variable

* Tx history window: removig always on top setting


interface/components/TxHistory/TxRow.js
```diff
@@ -12,14 +12,15 @@ class TxRow extends Component {
...
-      : new BigNumber(tx.value);
+      : new BigNumber(value || 0);
     const etherAmount = theValue
       .dividedBy(new BigNumber('1000000000000000000'))
       .toFixed();
     return etherAmount;
   };
 
-  toBigNumber = value => {
+  toBigNumber = v => {
```
marcgarreau: nonblocking: can use default params. `(value = 0) => { ... }`

VIP21:
- [ ]

PR #4165, opened 2018-12-30
Meteor update

etardz:


evertonfraga:
This branch hasn't yet performed as expected, which is to smoothly run AppVeyor built without using prebuilt `build-interface` directory.

I am closing it, but we aim to leave the branch sitting there.

